PR #3, opened 2022-01-19
'Solution'

denyslipin:


liuda-chuzdiuk (approved):

app/main.py
```diff
@@ -1,8 +1,19 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for one_person in people:
+        person_list.append(Person(one_person["name"], one_person["age"]))
+    for i in range(len(people)):
+        if "wife" in people[i] and None is not people[i]["wife"]:
```
IvanRamyk, line 15: It would be better to write second condition like this:
```
people[i]["wife"] is not None
```
It would be more readable

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.name = name
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for one_person in people:
```
IvanRamyk, line 12: It seems that 'one_' part is extra in the variable name. You can use just `person`

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #4, opened 2022-01-19
Solution

Abnormaltype:


denyslipin (approved):
It's great! I think your check "persons.get('..')" is perfect in this case.

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    pupil_array = [Person(person["name"], person["age"]) for person in people]
+
+    for persons in people:
```
IvanRamyk, line 13: Could you rename `persons` -> `person`. You store a single person in it.

IvanRamyk (changes requested):

app/main.py
Abnormaltype: okay)

IvanRamyk (approved):
Good job!

PR #9, opened 2022-01-20
'Solution'

serhii56465:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
+            except KeyError:
+                pass
+            try:
+                if j["husband"] is not None:
+                    Person.people[j["name"]].husband = \
+                        Person.people[j["husband"]]
+            except KeyError:
+                pass
+
+    return [Person.people[i] for i in Person.people]
```
Krimat, line 28: Second call of 'create_person_list' will return incorrect output

app/main.py
```diff
@@ -1,8 +1,28 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for i in people:
+        for key in i:
+            Person(i["name"], i["age"])
+    for j in people:
+        for key in j:
```
Krimat, line 15: Inner 'for' loop is redundant

Krimat (changes requested):

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for i in people:
+        for key in i:
```
IvanRamyk, line 12: It seems that this for-loop is extra here

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        except KeyError:
+            pass
+        try:
+            if j["husband"] is not None:
+                Person.people[j["name"]].husband = \
+                    Person.people[j["husband"]]
+        except KeyError:
+            pass
+    result = []
+    for i in range(len(people)):
```
IvanRamyk, line 27: You can remove this loop and append people to the `result` list when you create them (line 13)

app/main.py
```diff
@@ -1,8 +1,29 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for i in people:
+        for key in i:
+            Person(i["name"], i["age"])
+    for j in people:
+        try:
```
IvanRamyk, line 15: It would be better to check if key "wife" in dict using `if "wife" in j` and not to use try-except here.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for i in people:
+        for key in i:
+            Person(i["name"], i["age"])
+    for j in people:
```
IvanRamyk, line 14: Please, change the name of this variable to `person` or something more informative

IvanRamyk (changes requested):

IvanRamyk (changes requested):
Well done!
Waiting small fixes

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for i in range(len(people)):
+        Person(people[i]["name"], people[i]["age"])
```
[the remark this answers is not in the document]
serhii56465: I don't understand this comment. Could you explain in another way?

app/main.py
```diff
@@ -1,8 +1,24 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for i in range(len(people)):
+        result.append(
+            Person(
+                people[i]["name"], people[i]["age"]
+            ).people[people[i]["name"]]
```
IvanRamyk, line 16: You can add just a `Person` instance. You don't need to take `people` attribute from it
``` result.append(Person(....))```

**Review by IvanRamyk — CHANGES_REQUESTED** [review]

File: app/main.py (hunk 8 above)
Comment (serhii56465) [reply app/main.py:16]: I did it like this (see below), but you said that I should write in one line that's why I change to one line **result.append(Person(people[i]["name"],people[i]["age"]).people[people[i]["name"]])**:
PRIOR CODE:
   result = []
   for i in range(len(people)):
       Person(people[i]["name"], people[i]["age"])
       result.append(Person.people[people[i]["name"]])

Comment (serhii56465) [reply app/main.py]: I've done some changes. Could you look at it.

File: app/main.py [hunk 9]
```diff
@@ -1,8 +1,20 @@
...
+    def __init__(self, name, age):
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for i in range(len(people)):
```
IvanRamyk, line 11: You don't really need this cycle. You can create instances of `Person` class inside list comprehension  in line 13.
```
result = [Person(people[i]["name"], people[i]["age"]) for ... ]
``` 
In this case, you will create a new person and add it to the list in one command
If you still have questions about it you can ask in the `py_jan22_questions_and_answers` chat

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #10, opened 2022-01-20
Solution

alix-mak:


TPolina (approved):
I think your solution is good!

IvanRamyk (approved):
Well done!

PR #14, opened 2022-01-20
Solution

phantom-terminal:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    ppl_list = [Person(pers["name"], pers["age"]) for pers in people]
+
+    for pers in people:
+        if "wife" in pers and pers["wife"] is not None:
+            Person.people[pers["name"]].wife = Person.people[pers["wife"]]
+
+    for pers in people:
```
IgorOlifirenko, line 17: You can remove this loop and put this "if's" in one loop. And maybe better to use "if-elif ".

IgorOlifirenko, line 17: You can remove this loop and put this "if's" in one loop. And maybe better to use "if-elif ".

IgorOlifirenko (changes requested):

app/main.py
phantom-terminal: Removed unnecessary for loop. Tnx)

IgorOlifirenko (approved):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    def __init__(self, name, age):
+        self.age = age
+        self.name = name
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    ppl_list = [Person(pers["name"], pers["age"]) for pers in people]
```
IvanRamyk, line 11: Please, Do not shorten the names of the variables. It could impair the readability of the code.
It's not clear that `ppl_list` stores people. 

The same case with the `pers` variable.

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #18, opened 2022-01-20
Solution

Nazar-kob:


app/main.py
```diff
@@ -1,8 +1,17 @@
...
+        self.name = name
+        self.age = age
+        Person.people.update({name: self})
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = [Person(i['name'], i['age']) for i in people]
+    for j in people:
```
IvanRamyk, line 12: Could you use more descriptive names for variables?
You can use `person` instead of `i` and `j`

IvanRamyk (changes requested):

app/main.py
Nazar-kob: Yes, you're right

liuda-chuzdiuk (approved):

IvanRamyk (approved):
Well done!

PR #19, opened 2022-01-20
Solution

nikon4uk:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = [Person(person["name"], person["age"]) for person in people]
+    for person in people:
+        wife = person.get('wife')
+        husband = person.get('husband')
```
roland813, line 15: Іn my opinion it is not necessary to create a variable if you use once

roland813 (approved):

app/main.py
IvanRamyk: It's fine, especially when it improves readability of the code

IvanRamyk (approved):
Well done!

PR #20, opened 2022-01-20
Solution

NikitaZH1812:


8Men5Camels (approved):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = [Person(singel_person["name"], singel_person["age"])
+                   for singel_person in people]
+
+    for singel_person in people:
```
IvanRamyk, line 14: Typo in word `single`
Also It'd be better to use just `person` as a name for this variable, because word `single` doesn't tell us anything new about the variable. It's clear that only one person is stored in the variable `person`

IvanRamyk (changes requested):
God job!
Waiting for small fixes

IvanRamyk (approved):
Good job!

PR #23, opened 2022-01-21
'Solution'

Oleksandr563487953:


vldtsw (approved):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for item in people:
```
IvanRamyk, line 12: `item` doesn't tell anything about the variable. Choose the name which describes the data is stored, please

IvanRamyk (changes requested):

app/main.py
Oleksandr563487953: Cycle variables changed.

MasterpieceElbow (approved):

PR #25, opened 2022-03-24
Solution . Katiukha

KatiukhaO:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
-    # write your code here
-    pass
+
+def create_person_list(peoples: list) -> list:
+    person_list = [Person(person["name"], person["age"]) for person in peoples]
+    for person in peoples:
+        if "wife" in person and person["wife"]:
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        elif "husband" in person and \
+                person["husband"]:
```
TPolina, line 16: this line is not too long to split it

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+            Person.people[person["name"]].husband = \
+                Person.people[person["husband"]]
```
TPolina, line 18: bad style, better make temporary variable `husband` or smth else

TPolina (changes requested):

app/main.py
KatiukhaO: fix it

app/main.py
KatiukhaO: ok, fix it.  have a questios, ask on q and a

TPolina (approved):
Well done!

PR #28, opened 2022-05-10
solution

MarkoKhodan:


app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
+        self.name = name
+        self.age = age
```
Abnormaltype, line 7: `Create class Person. It's constructor takes and store name, age of a person. This class also should have a class attribute people, it is a dict that stores Person instances by their name. Constructor should add elements to this attribute.`

Please implements this.

Abnormaltype (changes requested):

Abnormaltype (approved):

PR #30, opened 2022-05-14
Solution

YuriiKindrat:


app/main.py
```diff
@@ -1,8 +1,31 @@
...
+    result = []
+    for person in people:
+        name = person['name']
+        age = person['age']
+        if 'wife' in person:
+            wife = person['wife']
+            result.append(Person(name, age, wife=wife))
+        if 'husband' in person:
+            husband = person['husband']
+            result.append(Person(name, age, husband=husband))
```
Magus6784, line 24: You added "wife" and "husband" twice. Here and at $26 - $30. 
Also we dont need $7 - $10, because we add with Person.people[person['name']].husband which automatly added self.husband or self.wife

Magus6784 (changes requested):

app/main.py
```diff
@@ -1,8 +1,19 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
```
Magus6784, line 4: good practice
name: str, age: int

Magus6784 (approved):

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person['name'], person['age'])
```
TPolina, line 12: use double quotes everywhere

TPolina (changes requested):

TPolina (approved):

PR #31, opened 2022-05-15
'Solution'

DmytroSvirsa:


app/main.py
```diff
@@ -1,8 +1,29 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(
+            person["name"],
+            person["age"],
+            wife=person["wife"] if "wife" in person else None,
+            husband=person["husband"] if "husband" in person else None
```
Magus6784, line 20: You added "wife" and "husband" twice. Here and at $23 - $28.
Also we dont need $7 - $10, because we add with Person.people[person['name']].husband which automatly added self.husband or self.wife

Magus6784 (changes requested):

app/main.py
```diff
@@ -1,29 +1,29 @@
...
+
+    @classmethod
+    def is_wife(cls, person):
+        if "wife" in person and person["wife"] is not None:
+            if person["wife"] in cls.people:
+                cls.people[person["name"]].wife = cls.people[person["wife"]]
+                cls.people[person["wife"]].husband = cls.people[person["name"]]
+
+    @classmethod
+    def is_husband(cls, pers):
```
Magus6784, line 17: i think better use the same name like in wife check
And use 
person: dict 
Good practice too

app/main.py
```diff
@@ -1,29 +1,29 @@
...
-            husband=person["husband"] if "husband" in person else None
-        )
-    for person in people:
-        if "wife" in person and person["wife"] is not None:
-            Person.people[person["name"]].wife = \
-                Person.people[person["wife"]]
-        if "husband" in person and person["husband"] is not None:
-            Person.people[person["name"]].husband = \
-                Person.people[person["husband"]]
+        Person(person["name"], person["age"])
```
Magus6784, line 26: You can use one "for" for add people on Person.people and another one for add wife/husband, so you can delete $12, $14, $19, $21 but i dont think this is better than your version

Magus6784 (approved):

app/main.py
```diff
@@ -1,8 +1,29 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
TPolina, line 7: better not to change cls attribute through object even if this attribute is mutable

TPolina (changes requested):
Your solution is original, but you will significantly simplify your code if you don't use `is_wife` and `is_husband` methods, and do all checks and add husbands/wives inside `create_person_list` function. You can avoid some checks and double operations.

TPolina (approved):
Well done!

PR #35, opened 2022-05-16
'Solution'

Korenely:


BohdanKryven (approved):

app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
TPolina, line 7: better not to change cls attribute through object even if this attribute is mutable

app/main.py
```diff
@@ -1,8 +1,23 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    res = [
+        Person(person['name'], person['age'])
+        for person in people
+    ]
+
+    for i in range(len(res)):
```
TPolina, line 16: better iterate through people not indexes. this will be more readable

app/main.py
TPolina: not fixed yet

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(peoples: list) -> list:
+    resultList = [Person(person['name'], person['age']) for person in peoples]
```
TPolina, line 11: use snake case and double quotes everywhere

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(peoples: list) -> list:
```
TPolina, line 10: word `peoples` is incorrect

TPolina (changes requested):

TPolina (approved):
Good job!

PR #38, opened 2022-05-16
'Solution'

Aleksei-Isaev:


app/main.py
```diff
@@ -1,8 +1,38 @@
...
+            self.wife = wife
+
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    result_list = []
```
Magus6784, line 19: You dont need result_list and you never used him in your code

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+
+    for person in people:
+        name = person['name']
+        age = person['age']
+        if 'wife' in person:
+            wife = person['wife']
+            result_list.append(Person(name, age, wife=wife))
+        if 'husband' in person:
+            husband = person['husband']
+            result_list.append(Person(name, age, husband=husband))
```
Magus6784, line 29: You added "wife" and "husband" twice. Here and at $32 - $36
Also we dont need $9 - $12, because we add with Person.people[person['name']].husband which automatly added self.husband or self.wife

app/main.py
```diff
@@ -1,8 +1,38 @@
 class Person:
-    # write your code here
-    pass
+
+    people = {}
+
+    def __init__(self, name, age, wife=None, husband=None):
```
Magus6784, line 5: Add annotation is good practice

app/main.py
Aleksei-Isaev: > You dont need result_list and you never used him in your code

And in your opinion, what do I have to do?

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+        self.name = name
+        self.age = age
+
+        if husband is not None:
+            self.husband = husband
+        if wife is not None:
+            self.wife = wife
```
TPolina, line 12: you do double or even triple work. Let your constructor take just name and age. Then you can rewrite your create_person_list function and avoid some checks and attribute assigns.

app/main.py
```diff
@@ -1,8 +1,38 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    result_list = []
+
+    for person in people:
+        name = person['name']
```
TPolina, line 22: use double quotes everywhere

TPolina, line 22: use double quotes everywhere

TPolina (changes requested):

TPolina (approved):
Well done!

PR #39, opened 2022-05-16
'Solution'

darten360:


TPolina (changes requested):
fix flake8

app/main.py
```diff
@@ -1,8 +1,25 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
+
+    for person in people:
+        if 'husband' in person and person['husband'] is not None:
```
TPolina, line 15: use double quotes everywhere

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+            Person.people[
+                person["name"]
+            ].husband = Person.people[
+                person["husband"]
+            ]
```
TPolina, line 20: not really good long code splitting; better use intermediate variable

TPolina (changes requested):

TPolina (approved):
Good!

PR #42, opened 2022-05-17
py-person-class solution

Y-Havryliv:


app/main.py
```diff
@@ -1,8 +1,35 @@
...
+        Person(person["name"], person["age"])
+
+    for person in people:
+        our_object = Person.people[person["name"]]
+
+        if "wife" in person and person["wife"] is not None:
+            our_object.wife = Person.people[person["wife"]]
+            result.append(our_object)
+
+        elif "husband" in person and person["husband"] is not None:
```
KatiukhaO, line 28: is not None dont nessesary write, because -  person["husband"] not in dict  , condition dotn executed code

KatiukhaO (changes requested):

app/main.py
Y-Havryliv: but it can be in dict as None(e.g. {"name": "Rachel", age: 30, "husband": None})

KatiukhaO (approved):

KatiukhaO (approved):

app/main.py
```diff
@@ -1,8 +1,35 @@
...
+        if "wife" in person and person["wife"]:
+            our_object.wife = Person.people[person["wife"]]
+            result.append(our_object)
+
+        elif "husband" in person and person["husband"]:
+            our_object.husband = Person.people[person["husband"]]
+            result.append(our_object)
+
+        else:
+            result.append(our_object)
```
TPolina, line 33: better write `append` outside if-else statement so you don't need to write it three times

TPolina (changes requested):

TPolina (approved):
Well done!

PR #46, opened 2022-05-17
Solution

Kiawami:


app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
```
Magus6784, line 4: I love annotation!

Magus6784 (approved):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
Magus6784, line 7: Sorry, i forget abour this line
better not to change cls attribute through object even if this attribute is mutable
So fix this and all ok

Magus6784 (changes requested):

TPolina (approved):
Well done!

PR #48, opened 2022-05-17
'The_Soution'

SergeyDev1996:


app/main.py
```diff
@@ -1,8 +1,38 @@
...
+    def __init__(self, name, age):
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
     # write your code here
-    pass
+    print(len(people))
```
Y-Havryliv, line 12: get rid of this unnecessary print statement

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+    person_list = []
+    for person in people:
+        if "wife" in person and person['wife'] != None:
```
Y-Havryliv, line 15: != None is reduntant, you should just use person["wife"]

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+            name = Person(person["name"], person["age"])
+            name.wife = person['wife']
+
+            person_list.append(name)
+            continue
+        if "husband" in person and person['husband'] != None:
```
Y-Havryliv, line 21: != None is reduntant, you should just use person["husband"]

app/main.py
Y-Havryliv, line 15: use double quotes everywhere

app/main.py
```diff
@@ -1,8 +1,38 @@
...
     # write your code here
-    pass
+    print(len(people))
+    person_list = []
+    for person in people:
+        if "wife" in person and person['wife'] != None:
+            name = Person(person["name"], person["age"])
+            name.wife = person['wife']
+
+            person_list.append(name)
```
Y-Havryliv, line 19: you can write "append" outside if-else statement in order not to repeat it three times

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+            name.husband = person['husband']
+
+            person_list.append(name)
+            continue
+        else:
+            name = Person(person["name"], person["age"])
+            person_list.append(name)
+    for person in people:
+        persona = Person.people[person['name']]
+        if 'wife' in person and person['wife'] is not None:
```
Y-Havryliv, line 32: is not None is redundant

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+            persona.wife = Person.people[person['wife']]
+
+        if 'husband' in person and person['husband'] is not None:
```
Y-Havryliv, line 35: is not None is redundant

Y-Havryliv (changes requested):

app/main.py
SergeyDev1996: the code won't work in this way

lenasharabura: you don't need continue
del it and write "append" outside if-else statement in order not to repeat it three times
and your cod will work correct

lenasharabura (approved):

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        if "wife" in person and person["wife"]:
+            name = Person(person["name"], person["age"])
+            name.wife = person["wife"]
+
+        if "husband" in person and person["husband"]:
+            name = Person(person["name"], person["age"])
+            name.husband = person["husband"]
+        else:
+            name = Person(person["name"], person["age"])
+        person_list.append(name)
```
TPolina, line 22: you do redundant operations: double checks (Ex: `if "wife" in person and person["wife"]:`) and redundant assignments (Ex: `name.wife = person["wife"]` - you don't need it as `wife` should be of type `Person`, no need to assign something temporal).
Better write short first cycle where you just create instances of Person class and in the second cycle do what you do: check if wife/husband in person and assign her/him

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    for person in people:
+        persona = Person.people[person["name"]]
```
TPolina, line 24: don't use variable name `persona` it has different meening

TPolina (changes requested):

MasterpieceElbow (approved):

PR #49, opened 2022-05-18
Solution

RomaTsybulia:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(peoples: list) -> list:
+    persons = []
```
mens0, line 11: you can rewrite it by using list comprehension

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+    for person in peoples:
+        persons.append(Person(person['name'], person['age']))
+
+    for people in peoples:
```
mens0, line 15: maybe better to use something like "for person in people", because "people" here is perceived like something plural

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        if 'wife' in people and people['wife'] is not None:
```
mens0, line 16: expresion "is not None" is redundant, it's the same as True

mens0 (changes requested):

mens0 (approved):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    def __init__(self, name, age):
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = [Person(person['name'], person['age']) for person in people]
```
TPolina, line 11: please, choose another variable name. word `persons` is incorrect

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+
+    for personage in people:
```
TPolina, line 13: choose better variable name

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        if 'wife' in personage and personage['wife']:
```
TPolina, line 14: use double quotes everywhere

TPolina (changes requested):

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+    list_of_people = [
+        Person(person["name"], person["age"])
+        for person in people
+    ]
+
+    for person in people:
+        if "wife" in person and person["wife"]:
+            wife = Person.people[person["wife"]]
+            Person.people[person["name"]].wife = wife
+        elif 'husband' in person and person["husband"]:
```
TPolina, line 20: use double quotes

TPolina (changes requested):

TPolina (approved):
Nice!

PR #52, opened 2022-05-18
Solution

NOSSUDAK:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        age = person["age"]
+        people_list.append(Person(name, age))
+    for person in people:
+        name = person["name"]
+        if person.get('wife') is None and person.get('husband') is None:
+            continue
+        elif person.get('wife') is not None:
+            Person.people[name].wife = Person.people[person["wife"]]
+        else:
+            Person.people[name].husband = Person.people[person["husband"]]
```
Magus6784, line 23: You can rewrite in 4 lines with double check
`if "wife" in human and human["wife"] is not None:
            Person.people[human["name"]].wife = Person.people[human["wife"]]`
And same thing for husband
So you dont need 'if' with continue

Magus6784 (changes requested):

app/main.py
```diff
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
```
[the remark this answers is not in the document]
NOSSUDAK: You are right, I will add.

app/main.py
```diff
@@ -1,8 +1,24 @@
...
-    # write your code here
-    pass
+    people_list = []
+    for person in people:
+        name = person["name"]
+        age = person["age"]
+        people_list.append(Person(name, age))
+    for person in people:
+        name = person["name"]
+        if person.get('wife') is None and person.get('husband') is None:
```
TPolina, line 18: use double quotes everywhere

app/main.py
TPolina: agree with Maksym, better repeat lines 20-21 for husband and you will not need the first check

TPolina (changes requested):

PR #53, opened 2022-05-18
'Solution'

dashaprotsko:


Korenely (approved):

app/main.py
```diff
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
TPolina, line 7: better not to change class attribute through object even if this attribute is mutable

app/main.py
```diff
@@ -1,8 +1,22 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = [
+        Person(person['name'], person['age'])
+        for person in people
+    ]
+
+    for el in range(len(result)):
```
TPolina, line 16: better iterate through people not indexes. This will be more readable

TPolina (changes requested):

TPolina (approved):
Well done!

PR #56, opened 2022-05-18
Solution

Chardona:


app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
DmytroSvirsa, line 7: better not to change `cls` attribute through object even if this attribute is mutable

app/main.py
```diff
@@ -1,8 +1,20 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    friends = []
+    for hu_man in people:
+        friend = Person(hu_man["name"], hu_man["age"])
+        friends.append(friend)
+    for pos, hu_man in enumerate(people):
```
DmytroSvirsa, line 15: hu_man incorrect variable name

DmytroSvirsa (changes requested):

app/main.py
Chardona: You didn't get the joke.

app/main.py
Chardona: suggestions how  else to add elements to the dictionary on initialization?

Magus6784: `        Person.people[self.name] = self`

Magus6784 (changes requested):
Using annotation very good practice))
(just fix $7, annotations are my thing)

Magus6784 (approved):

PR #59, opened 2022-05-18
Solution

vasylhnatiuk:


app/main.py
```diff
@@ -1,8 +1,35 @@
...
+                person.husband = person.partner
+                people_with_partner.append(person)
+        new_people_list.append(person)
+
+    for person1 in people_with_partner:
+        person2 = person1.people[person1.partner]
+        if 'wife' in person1.__dict__:
+            person1.wife = person2
+        elif 'husband' in person1.__dict__:
+            person1.husband = person2
```
Magus6784, line 34: You have duplicate code here, because you're double times add wife/husband
I think better split add to Person.people and check wife/husband with 2 for
one for         Person(human["name"], human["age"])
second for add wife/husband
Good Luck!

Magus6784 (changes requested):

app/main.py
vasylhnatiuk: thanks)

app/main.py
```diff
@@ -1,8 +1,28 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
TPolina, line 7: better not to change class attribute through instance even if this attribute of a mutable type

app/main.py
```diff
@@ -1,8 +1,28 @@
...
-    # write your code here
-    pass
+    new_people = []
+    for person_dict in people:
+        person = Person(person_dict['name'], person_dict['age'])
+        new_people.append(person)
+
+    for index in range(len(new_people)):
+        if 'wife' in people[index].keys():
+            if people[index]['wife']:
```
TPolina, line 18: you can write two `if` statements in one line using `and` operator

TPolina (changes requested):

TPolina (changes requested):

app/main.py
```diff
@@ -1,8 +1,28 @@
...
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
+        self.name = name
+        self.age = age
+
+    def add_person(self):
+        self.people[self.name] = self
```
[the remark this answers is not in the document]
vasylhnatiuk: ok, thanks)

TPolina (approved):

PR #68, opened 2022-06-06
Solution

andrii-padalko:


app/main.py
```diff
@@ -1,8 +1,17 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    lst = [Person(people_["name"], people_["age"]) for people_ in people]
+    for i, people_ in enumerate(people):
+        if "wife" in people_ and people_["wife"] is not None:
+            lst[i].wife = Person.people[people_["wife"]]
+        if people_.get("husband") and people_["husband"] is not None:
```
IvanRamyk, line 15:
```suggestion
        if "husband" in people_ and people_["husband"] is not None:
```

IvanRamyk (approved):
Let's use the same style for both conditions

Abnormaltype (approved):

PR #80, opened 2022-06-22
Solution_v1

Vrungels:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_person = []
+
+    for one_dict in people:
```
IvanRamyk, line 13: `one_dict` isn't a descriptive variable name 
You can use `person_dict` for instance

IvanRamyk (changes requested):

app/main.py
IvanRamyk, line 13: Please, change the name of the variable here too

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #83, opened 2022-06-24
Solution

InnaKushnir:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    if people:
```
IvanRamyk, line 12: Why use this if? 
It makes the code more complicated

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        for member in people:
+            person_list.append(Person(member["name"], member["age"]))
+
+        for member in people:
+            if "husband" in member.keys():
+                if member["husband"] is not None:
```
IvanRamyk, line 18: Avoid using nested ifs. It's better to use `and` operator

app/main.py
```diff
@@ -1,8 +1,27 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    if people:
+        for member in people:
+            person_list.append(Person(member["name"], member["age"]))
+
+        for member in people:
+            if "husband" in member.keys():
```
IvanRamyk, line 17:
```suggestion
            if "husband" in member:
```
It's more clear way to check key presence

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #84, opened 2022-06-27
'Solution'

MilManIT:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = [Person(human["name"], human["age"]) for human in people]
+
+    for i in people:
```
Abnormaltype, line 14: what is `i`?

Abnormaltype (changes requested):

app/main.py
MilManIT: my bad!)

Abnormaltype (approved):
well done)

app/main.py
Abnormaltype: ))

PR #91, opened 2022-07-06
Solution

DHushchyk:


app/main.py
```diff
@@ -1,8 +1,35 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    my_list = []
```
IvanRamyk, line 13: Why is it yours?)

IvanRamyk: Please, choose another name

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,8 +1,35 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    info_list = []
+
+    for person in people:
+        Person(person["name"], person["age"])
```
IvanRamyk, line 16: Could you add the created person to the list in this line 
It would help to avoid code duplication

IvanRamyk (changes requested):

app/main.py
DHushchyk: I supposed I must add new instances before adding persons to the list, but it was yesterday) Shorted cod will be better. Thanks!

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+        info_list.append(Person.people[person["name"]])
```
IvanRamyk, line 17: You create duplicates in this line

IvanRamyk (approved):
Well done!

Abnormaltype (approved):

PR #92, opened 2022-07-10
Solution

shurushku:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person['name'], person['age']) for person in people]
+
+    for person in people:
+        if 'wife' in person.keys() and person['wife']:
```
MasterpieceElbow, line 14: You shouldn't use `.keys()` to check if there is a key.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #103, opened 2022-07-27
'Solution'

Pavilion4ik:


Abnormaltype (approved):
Well done!

Pavilion4ik:
Вітаю. На платформі задача не позначена як виконана, хоча виконана успішно вже давненько.

app/main.py
```diff
@@ -1,8 +1,18 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name, age):
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
```
Bohdan-Salamakha, line 10: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/main.py
```diff
@@ -1,8 +1,18 @@
...
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"])
+                   for person in people]
+    for hum in people:
+        if "wife" in hum and hum["wife"]:
+            Person.people[hum["name"]].wife = Person.people[hum["wife"]]
+        if "husband" in hum and hum["husband"]:
+            Person.people[hum["name"]].husband = Person.people[hum["husband"]]
```
Bohdan-Salamakha, line 17: use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (changes requested):
fix tests

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(peo["name"], peo["age"]) for peo in people]
+    for hus in people:
+        if "wife" in hus and hus.get("wife"):
```
ArsenPidhoretskyi, line 13: You can simplify this: just remove "wife" in hus
The same with husband
```suggestion
        if hus.get("wife"):
```

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(peo["name"], peo["age"]) for peo in people]
```
ArsenPidhoretskyi, line 11: peo is not clear name rename it to person

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+    for hus in people:
```
ArsenPidhoretskyi, line 12: rename hus to person

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        if "wife" in hus and hus.get("wife"):
+            Person.people.get(hus.get("name")).wife = \
+                Person.people[hus.get("wife")]
```
ArsenPidhoretskyi, line 15:
```suggestion
            Person.people[hus["name"]].wife = Person.people[hus["wife"]]
```
In this case not need .get because we already know that wife exist in hus
Also we know that we already create hus with name
The same fix with husband

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: Here it no needed

app/main.py
ArsenPidhoretskyi: Because .get can return None or you must provide default value, but we no have default value in this case
https://docs.quantifiedcode.com/python-anti-patterns/correctness/not_using_get_to_return_a_default_value_from_a_dictionary.html?highlight=get

app/main.py
ArsenPidhoretskyi: Because:
1. We already know that "husband" exist in hum
2. And we know that "name" exist also, due to task requirements
https://docs.quantifiedcode.com/python-anti-patterns/correctness/not_using_get_to_return_a_default_value_from_a_dictionary.html?highlight=get

PR #124, opened 2022-08-26
Solution

hai-le-di:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result_list = []
+    for person in people:
+        result_list.append(Person(person["name"], person["age"]))
+    for obj in result_list:
```
IvanRamyk, line 15: `obj` isn't a good variable name, please, change it

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        for person in people:
```
IvanRamyk, line 16: You don't need nested cycles in this task, try to simplify an algorithm

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result_list = []
+    for person in people:
+        item = Person(person["name"], person["age"])
```
IvanRamyk, line 14:
```suggestion
        person = Person(person["name"], person["age"])
```

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result_list = []
+    for person in people:
```
IvanRamyk, line 13:
```suggestion
    for person_dict in people:
```

app/main.py
IvanRamyk: `item` isn't a descriptive name as well
I suggest you use the `person` variable to store a `Person` instance and `person_dict` to store a dictionary with information about the person you are going to create

IvanRamyk (changes requested):

DmytroSvirsa (approved):

PR #132, opened 2022-09-02
'SALUT'

SorokaIgor:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_list = []
+    for i in people:
+        new_list.append(Person(i['name'], i['age']))
+
+    for x in people:
+        if 'wife' in x.keys():
```
IvanRamyk, line 16:
```suggestion
        if 'wife' in x:
```

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+            if x['wife'] is not None:
```
IvanRamyk, line 17: use `and` operator instead of nested `if`s

IvanRamyk (changes requested):

app/main.py
SorokaIgor: done

app/main.py
SorokaIgor: done

IvanRamyk (approved):
Well done!

PR #135, opened 2022-09-06
Solution

vlad-milkov:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+        if person.get("wife"):
+            new_person.wife = person["wife"]
+        elif person.get("husband"):
+            new_person.husband = person["husband"]
```
DmytroSvirsa, line 17: You can do it without these checks and additions

DmytroSvirsa (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+    person_list = []
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+        person_list.append(new_person)
+
+    for ind in people:
+        if ind.get("wife"):
+            Person.people[ind["name"]].wife = Person.people[ind["wife"]]
+        elif ind.get("husband"):
+            Person.people[ind["name"]].husband = Person.people[ind["husband"]]
```
DmytroSvirsa, line 20: `ind` is not enough describable name, and let's do it with `if` instead `elif`

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #154, opened 2022-09-16
Develop

musakovskiy:


app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list):
+    new_list = []
```
IvanRamyk, line 11: Choose a more descriptive name for the list

IvanRamyk (changes requested):

app/main.py
musakovskiy: name list changed

PR #155, opened 2022-09-16
Solution)

Gr1zzZzzly:


mltdll (approved):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    for per in people:
```
TPolina, line 14: per is a bad variable name

TPolina (changes requested):

app/main.py
Gr1zzZzzly: yep but the line has been too long with "person"

TPolina (approved):

PR #158, opened 2022-09-18
Solution

Aleksandro777:


app/main.py
```diff
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
olia-trofymchuk, line 7: it is better to use the class name to get access to the class attributes.
you can read about it here https://github.com/mate-academy/py-herbivores-and-carnivores/blob/master/checklist.md

olia-trofymchuk (changes requested):

olia-trofymchuk (approved):

TPolina (approved):
Well done!

PR #159, opened 2022-09-18
Solution<3

Mowka7:


app/main.py
```diff
@@ -1,8 +1,25 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
wQuelS, line 7: It is better to use the class name to get access to the class attributes.
you can read about it here [https://github.com/mate-academy/py-herbivores-and-carnivores/blob/master/checklist.md](url)

wQuelS (changes requested):

wQuelS (approved):
looks good imo

PR #161, opened 2022-09-18
Solution

skrypka-dmytro:


app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
```
wQuelS, line 4: add type annotations

app/main.py
```diff
@@ -1,8 +1,23 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for human in people:
+        person_list.append(Person(human["name"], human["age"]))
+
+    for human in people:
+        if "wife" in human and human["wife"] is not None:
```
wQuelS, line 17: consider of using get() instead of checking if key in dictionary

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
+        elif "husband" in human and human["husband"] is not None:
```
wQuelS, line 19: same here

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
```
Y-Havryliv, line 11: btw you can use list comprehension here

Y-Havryliv (approved):

PR #162, opened 2022-09-19
Solution for the task

olia-trofymchuk:


app/main.py
```diff
@@ -1,8 +1,27 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
wQuelS, line 7: It is better to use the class name to get access to the class attributes.
you can read about it here https://github.com/mate-academy/py-herbivores-and-carnivores/blob/master/checklist.md

app/main.py
```diff
@@ -1,8 +1,27 @@
...
-    # write your code here
-    pass
+    list_with_people = []
+
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+        list_with_people.append(new_person)
+
+    for index, hum in enumerate(people):
+        if "wife" in hum:
```
wQuelS, line 18: consider of using get() instead of checking if key is in dictionary

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+            name = hum["wife"]
+            if name is not None:
+                list_with_people[index].wife = Person.people[name]
+        if "husband" in hum:
```
wQuelS, line 22: same here

wQuelS (changes requested):

wQuelS (approved):
Looks good imo:)

Y-Havryliv (approved):
Well done!

PR #164, opened 2022-09-19
Solution<3

Mowka7:


hasslieb3 (approved):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
-    pass
+    list_of_humans = []
+
+    for person in people:
+        new_person = Person(person['name'], person['age'])
+        list_of_humans.append(new_person)
+
+    for index, human in enumerate(people):
+        if 'wife' in human:
+            if human['wife'] is not None:
```
TPolina, line 19: it's better to use `and` instead of nested `if` blocks

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_humans = []
+
+    for person in people:
+        new_person = Person(person['name'], person['age'])
```
TPolina, line 14: please, use double quotes everywhere

TPolina (changes requested):

TPolina (approved):
Well done!

PR #165, opened 2022-09-19
Develop

KatrinLazarenko:


app/main.py
```diff
@@ -1,8 +1,36 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    res = []
+    i = 0
+    for person_data in people:
+        person = Person(person_data["name"], person_data["age"])
+        res.append(person)
+    for person_data in people:
```
mltdll, line 24: If you want to have access to indices while iterating through an array, you can use [`enumerate()`](https://docs.python.org/3/library/functions.html#enumerate) function:
```python
for i, person_data in enumerate(people):
...
```

mltdll (approved):

app/main.py
```diff
@@ -17,11 +17,10 @@ def get_person(cls, name: str):
 
 def create_person_list(people: list) -> list:
     res = []
-    i = 0
     for person_data in people:
         person = Person(person_data["name"], person_data["age"])
         res.append(person)
-    for person_data in people:
+    for i, person_data in enumerate(people):
         try:
```
mltdll, line 24: By the way, I failed to notice something else: you don't need `try` ... `except` blocks here. You can use `dict.get(key)` method: it returns None if key doesn't exist.

app/main.py
```diff
@@ -1,8 +1,28 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        self.add_person(self)
+
+    @classmethod
+    def add_person(cls, person: "Person"):
```
TPolina, line 10: `person: Person `is a correct type annotation

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        cls.people[person.name] = person
```
TPolina, line 11: Why don't you add object to dict directly in init?

TPolina (changes requested):

TPolina (approved):
Good job

PR #166, opened 2022-09-19
Solution

Kitazuka:


illiariaboy94 (approved):
Clear code, looks fine.

app/main.py
```diff
@@ -1,8 +1,24 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        person_list.append(Person(person["name"], person["age"]))
+    for index, person in enumerate(people):
+        if "wife" in person:
+            if person["wife"] is None:
+                continue
```
TPolina, line 18: іt's better to use `and` instead of nested `if` blocks

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+            person_list[index].wife = Person.people[person["wife"]]
+        else:
+            if person["husband"] is None:
+                continue
```
TPolina, line 22: same here, better: `if "husband" in person and person["husband"] is not None:`

TPolina (changes requested):

TPolina (approved):

PR #168, opened 2022-09-19
Solution Vadim Sabadyr

VadimSabadyr:


app/main.py
```diff
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
```
Andriy-Makarenko, line 4: Best practice is to add type annotation for constructor as well, not only for arguments. -> None

Andriy-Makarenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
+    for person in people:
+        if "husband" in person:
```
Y-Havryliv, line 14: it's better to use `and` instead of nested `if` blocks

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+            if person["husband"] in Person.people:
+                Person.people[person["name"]].husband \
+                    = Person.people[person["husband"]]
+
+        if "wife" in person:
```
Y-Havryliv, line 19: it's better to use `and` instead of nested `if` blocks

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+            if person["wife"] in Person.people:
+                Person.people[person["name"]].wife\
```
Y-Havryliv, line 21: don't use this `\=` , make a temporal variable instead

app/main.py
```diff
@@ -1,8 +1,24 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
+    for person in people:
+        if "husband" in person:
+            if person["husband"] in Person.people:
+                Person.people[person["name"]].husband \
```
Y-Havryliv, line 16: don't use this `\=` , make a temporal variable instead

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

Andriy-Makarenko (approved):

PR #170, opened 2022-09-19
'Solution'

hasslieb3:


app/main.py
```diff
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
```
HStakhiv, line 4: Please, add a type annotation

HStakhiv (changes requested):
Please, add a type annotation

app/main.py
```diff
@@ -1,8 +1,22 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    classes_people = [
+        Person(person["name"], person["age"])
+        for person in people
+    ]
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife =\
```
TPolina, line 17: don't use `=\` , create a temporal variable instead

app/main.py
TPolina: +

TPolina (changes requested):

app/main.py
TPolina: :heavy_plus_sign:

HStakhiv (approved):
Good

app/main.py
```diff
@@ -12,11 +12,9 @@ def create_person_list(people: list) -> list:
...
         for person in people
     ]
-    for person in people:
-        if person.get("wife"):
-            Person.people[person["name"]].wife =\
-                Person.people[person["wife"]]
-        elif person.get("husband"):
-            Person.people[person["name"]].husband =\
-                Person.people[person["husband"]]
+    for man in people:
```
TPolina, line 15: no, `man` is not a good variable name here. Better create intermediate variables wife/husband, get needed person and write it in this variables and then use them

TPolina (changes requested):

TPolina (approved):
Well done!

PR #181, opened 2022-09-20
Solution

misha-doroshenko:


app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
```
linaSam, line 4: add annotation for this function

linaSam (changes requested):

linaSam (approved):
Great :)

app/main.py
```diff
@@ -1,8 +1,20 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for i in range(len(people)):
+        if "wife" in people[i].keys():
+            if people[i]["wife"]:
```
TPolina, line 15: it's better to use `and` instead of nested `if` blocks

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+                person_list[i].wife = Person.people[people[i]["wife"]]
+        if "husband" in people[i].keys():
+            if people[i]["husband"]:
```
TPolina, line 18: same here

TPolina (changes requested):

TPolina (approved):

PR #182, opened 2022-09-20
solved the task

Cerne13:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```
wQuelS, line 15: consider to use get() instead of checking if key in dictionary

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+            Person.people[person["name"]].wife = \
```
wQuelS, line 16: consider to use temporary variable to get rid of =\. It is up to you, but mentor can reject this

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+                Person.people[person["wife"]]
+
+        elif "husband" in person and person["husband"] is not None:
+            Person.people[person["name"]].husband = \
```
wQuelS, line 20: same here

wQuelS (changes requested):

app/main.py
Cerne13: Can't remember if we used getters/setters in the course yet. So did w/o one. But yeah, it's better with a getter use. Thank you

app/main.py
Cerne13: Thanks! Yes, since there's no block scope in Python, it'd be much tidier with some minimal refactoring.

app/main.py
wQuelS: i meant dictionary.get(), so u don't have to check if key in dictionary

wQuelS (approved):
Looks great to me

TPolina (approved):
Well done!

PR #185, opened 2022-09-20
solution

pavlo-kapitanets:


app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
```
HStakhiv, line 7: It is better to use `self__class__.people[name]`

app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
```
HStakhiv, line 4: Please, add type anotation

app/main.py
```diff
@@ -1,8 +1,20 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_list = [Person(human["name"], human["age"]) for human in people]
+
+    for human in people:
+        if "wife" in human and human["wife"] is not None:
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
+        elif "husband" in human and \
```
HStakhiv, line 16: No need for `elif`. Please, change to `else`.

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+                human["husband"] is not None:
+            Person.people[human["name"]].husband \
```
HStakhiv, line 18: It is better to create a new variable that will contain `Person.people[human["name"]]`  and delete `\`

app/main.py
```diff
@@ -1,8 +1,20 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_list = [Person(human["name"], human["age"]) for human in people]
+
+    for human in people:
+        if "wife" in human and human["wife"] is not None:
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
```
HStakhiv, line 15: You can use the variable for `Person.people[human["name"]]`

HStakhiv (changes requested):
Please, make corrections

app/main.py
HStakhiv: Sorry, I did not notice the next `elif`

app/main.py
pavlo-kapitanets: All done

HStakhiv (approved):
Looks good

app/main.py
TPolina: @HStakhiv @pavlo-kapitanets 
No, you do unnecessary actions using `self.__class__`, and readability is much worse. To access class attributes we should use the class name. We can also use `cls`, but only when it's a classmethod.

app/main.py
```diff
@@ -1,8 +1,18 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
TPolina, line 7: You should access class attribute through class name

TPolina (changes requested):

app/main.py
HStakhiv: Isn't that "hard code"? And I saw an example of such solution in one of the mentor's solutions on the platform

TPolina: In this task `Person.people[self.name] = self` is better because you are not going to inherit this class.
> Explicit is better than implicit.

TPolina (approved):

PR #189, opened 2022-09-20
'Solution'

LiudmylaKulinchenko:


app/main.py
```diff
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
olia-trofymchuk, line 7: it's better to use Person.people instead of self

olia-trofymchuk (changes requested):

olia-trofymchuk (approved):
imo it looks great :)

TPolina (approved):
Well done!

PR #190, opened 2022-09-20
Solution

Qusert17:


app/main.py
```diff
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
linaSam, line 7: you can change self.__class__. for just name of class

linaSam (changes requested):

app/main.py
Qusert17: Done

linaSam (approved):
Great!

TPolina (approved):
Well done!

PR #192, opened 2022-09-20
Solution

dmytro-petliovanyi:


app/main.py
```diff
@@ -1,8 +1,27 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
Y-Havryliv, line 8: it's better to use class name while getting access to class attributes: `Person.people...`

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Nice!

PR #196, opened 2022-09-20
solution

FranchukS:


Kitazuka (approved):
I think looks good

app/main.py
```diff
@@ -1,8 +1,31 @@
 class Person:
-    # write your code here
-    pass
+
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
```
TPolina, line 8: you should access class attribute through class name

app/main.py
```diff
@@ -1,8 +1,31 @@
...
-    # write your code here
-    pass
+    person_list = []
+
+    for individual in people:
+        person = Person(individual["name"], individual["age"])
+        person_list.append(person)
+
+    for individual in people:
+        if "wife" in individual.keys() and \
```
TPolina, line 19: please, don't use `\ `, instead you can create a shorter variable name (person for ex) and skip keys() method (`if "wife" in individual` is the same)

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+                individual["wife"] is not None:
+
+            Person.people[individual["name"]].wife = \
```
TPolina, line 22: same here (create intermediate variable) and further

TPolina (changes requested):

app/main.py
FranchukS: Changed in "Final cut" commit

app/main.py
FranchukS: Used tips and rewrote

TPolina (approved):

PR #199, opened 2022-09-21
Solution

mariia-movchan:


FranchukS (approved):
Nice!

app/main.py
```diff
@@ -1,8 +1,21 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = \
```
TPolina, line 15: Don’t use `= \` , create temporal variable instead

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+                Person.people[person["wife"]]
+        elif person.get("husband"):
+            Person.people[person["name"]].husband = \
```
TPolina, line 18: same here

TPolina (changes requested):

app/main.py
mariia-movchan: fixed

TPolina (approved):
Nice work!

PR #203, opened 2022-09-21
Person class first attempt

VitaliyBychevoy:


app/main.py
```diff
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people: dict = {}
+
+    def __init__(self, name: str, age: int):
```
linaSam, line 4: add annotation for this function pls

linaSam (changes requested):

linaSam (approved):
Great!

app/main.py
```diff
@@ -1,8 +1,22 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for human in people:
+        person_list.append(Person(human["name"], human["age"]))
+
+    for human in people:
+        if human.get("husband") is not None:
+            Person.people[human["name"]].husband = \
```
TPolina, line 17: Don’t use `= \` , create temporal variable instead

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+                Person.people[human["husband"]]
+        if human.get("wife") is not None:
+            Person.people[human["name"]].wife = \
```
TPolina, line 20: Don’t use `= \` , create temporal variable instead

TPolina (changes requested):

TPolina (approved):

PR #235, opened 2022-10-11
'Solution'

Paul-Maslov:


app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for item in people:
```
IvanRamyk, line 12: `item` is a bad variable name

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        person = Person(item["name"], item["age"])
+        for key in item.keys():
```
IvanRamyk, line 14:
```suggestion
        for key in item:
```

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+                person.wife = item["wife"]
+                person.husband = None
+            if key == "husband":
+                person.husband = item["husband"]
+                person.wife = None
+        person_list.append(person)
+
+    for person in person_list:
+        if person.wife is not None:
+            for i in range(len(person_list)):
```
IvanRamyk, line 25: use `Person.people` dict to get rid of an extra cycle

IvanRamyk (changes requested):

app/main.py
Paul-Maslov: Sorry for the late reply. There were problems with electricity

IvanRamyk (approved):
Well done!

PR #243, opened 2022-10-24
Solutions

Foksanv:


inrock1 (approved):

app/main.py
```diff
@@ -1,8 +1,18 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    for name, person in Person.people.items():
+        for person_info in people:
+            if name == person_info["name"] and person_info.get("wife"):
+                person.wife = Person.people[person_info["wife"]]
+            if name == person_info["name"] and person_info.get("husband"):
+                person.husband = Person.people[person_info["husband"]]
```
YuriiKindrat, line 17: no need to check name here

YuriiKindrat (changes requested):

app/main.py
oksana-feshchenko: without this check pytest failing, so i will rewrite loops order

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,17 @@
...
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    for inf in people:
```
YuriiKindrat, line 12: inf is a bad name for variable, use more descriptive naming

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #244, opened 2022-10-25
'Solution'

Sebshe:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
+
+    for person in Person.people.values():
+        for this_person in people:
+            if person.name == this_person["name"]:
+                if "husband" in this_person:
+                    if this_person["husband"] is not None:
+                        person.husband = Person.people[this_person["husband"]]
+                elif "wife" in this_person:
+                    if this_person["wife"] is not None:
+                        person.wife = Person.people[this_person["wife"]]
```
Abnormaltype, line 25: too complicated, you can use here only one loop and use only one condition. Try to simplify this

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #304, opened 2022-12-17
'Solution'

timyr4iq:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        person_list.append(Person(person["name"], person["age"]))
+    for person in people:
+        for key in person:
```
katevasilenko, line 15: You don’t need this for loop. Your code looks too complicated now

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+            if key == "wife" and person["wife"] is not None:
```
katevasilenko, line 16: `If “wife” in person` Is much better to understand and helps you avoid useless loops

katevasilenko (changes requested):

PR #329, opened 2023-01-08
Solution

YuriiVataschuk:


app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.age = age
+        self.name = name
+        Person.people[f"{self.name}"] = self
```
katevasilenko, line 7: `Person.people[name] = self` would be enough

app/main.py
```diff
@@ -1,8 +1,23 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result_list = []
```
katevasilenko, line 11: Maybe `people_list` or something like that?

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+    for person in people:
+        result_list.append(Person(name=person["name"], age=person["age"]))
+    for person in people:
+        if "wife" in person.keys():
```
katevasilenko, line 15: You don’t need `.keys()`. It looks to keys by default

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+            if person["wife"] is not None:
+                person_wife = Person.people[person["wife"]]
+                Person.people[person["name"]].wife = person_wife
+        if "husband" in person.keys():
```
katevasilenko, line 19: Remove `.keys()`

katevasilenko (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #330, opened 2023-01-08
'Solution'

Vitalii2906:


app/main.py
```diff
@@ -1,8 +1,29 @@
 class Person:
-    # write your code here
-    pass
+    people = dict()
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
```
Vasyl-Smutok, line 7: it is better to write it like this
`Person.people[self.name] = self`

app/main.py
```diff
@@ -1,8 +1,29 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_list = list()
```
Vasyl-Smutok, line 11: new list of what?🧐
better change the variable name

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    for person in people:
+        if None in person.values():
+            new_list.append(Person(person["name"], person["age"]))
+        else:
+            keys = list(person.keys())
```
Vasyl-Smutok, line 16: `person.keys()` == `person`

therefore it is not necessary to use `keys()` here

Vasyl-Smutok (changes requested):

olena-hul (approved):

PR #331, opened 2023-01-08
Solution

oleksandr-lysanets:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = []
+    for person in people:
+        people_list.append(Person(person["name"], person["age"]))
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```
Serhii-Leonenko, line 16: You can use `if person.get("wife") is not None` if there is no such key dict.get() will return us a default value None

Serhii-Leonenko (changes requested):
Good

YuriiKindrat (approved):

PR #334, opened 2023-01-13
Solution

Rename925:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons_list = [Person(per["name"], per["age"]) for per in people]
```
Vasyl-Smutok, line 11: `per` it's bad variable name

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons_list = [Person(single_person["name"], single_person["age"])
+                    for single_person in people]
```
ArsenPidhoretskyi, line 12: person is already mean one person)

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```
ArsenPidhoretskyi, line 14: use dict.get method

app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
```
ArsenPidhoretskyi, line 2: Add type annotations
HINT: `from __future__ import annotations`

ArsenPidhoretskyi (changes requested):

app/main.py
Vasyl-Smutok, line 2: use 
`people: dict[str, Person] = {}`

Vasyl-Smutok (changes requested):

app/main.py
Rename925: я раніше не зустрічав такого запису, і я не розумію де я його маю використати

olena-hul: It is needed to provide type annotation for people attribute. You can change the second line as mentors are targeting it. But don't import annotations, as it will cause tests to fail. Use

people: dict[str, 'Person'] = {}

olena-hul (changes requested):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons_list = [Person(info["name"], info["age"])
```
olena-hul, line 11: We can't say `persons`, as person has no plural form. Use `people_list`

olena-hul (changes requested):

ArsenPidhoretskyi (approved):

PR #339, opened 2023-01-21
'Solution'

iishchenko:


app/main.py
```diff
@@ -1,8 +1,45 @@
...
+        if husband:
+            self.husband = husband
+
+
+def create_person_list(people1: list) -> list:
+    person_list1 = []
+    for person in people1:
+        one_person = Person(person["name"], person["age"])
+        one_person.add_wife(person.get("wife", None))
+        one_person.add_husband(person.get("husband", None))
```
ArsenPidhoretskyi, line 28: We need to add an attribute if the key exists and not None, not need to use the method in this way. Methods/setters will be needed if you fields will be private or protected

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+
+    def add_husband(self, husband: object) -> None:
+        if husband:
+            self.husband = husband
+
+
+def create_person_list(people1: list) -> list:
+    person_list1 = []
+    for person in people1:
+        one_person = Person(person["name"], person["age"])
```
ArsenPidhoretskyi, line 26: person is already one. You can use list comprehensions here

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+                        person.wife = object1
+        except AttributeError:
+            pass
+        try:
+            if person.husband:
+                for object1 in person_list1:
+                    if object1.name == person.husband:
+                        person.husband = object1
+        except AttributeError:
+            pass
```
ArsenPidhoretskyi, line 44: Just iterate in original list and check for [wife] and [husband] and that set, use Person.people

ArsenPidhoretskyi (changes requested):

iishchenko:
Changed:
Not necessary to use .update, just self.people[self.name] = self
bad naming, rename to people: list[Person]
We need to add an attribute if the key exists and not None, not need to use the method in this way. Methods/setters will be needed if you fields will be private or protected
person is already one. You can use list comprehensions here
Just iterate in original list and check for [wife] and [husband] and that set, use Person.people

app/main.py
```diff
@@ -1,8 +1,33 @@
 class Person:
-    # write your code here
-    pass
 
+    people: dict[str] = {}
```
ArsenPidhoretskyi, line 3: dict expect two params
dict[str, Person]

app/main.py
```diff
@@ -1,8 +1,33 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
+
+    def add_wife(self, wife: str) -> None:
+        if wife is not None:
```
ArsenPidhoretskyi, line 11: due to your type annotations it does not possible 
update typing: wife: Optional[str]

The same with husband

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+def create_person_list(
+        people: list[dict[str, str | int | None]]
+) -> list[Person]:
+    person_list = [
+        Person(person["name"], person["age"])
+        for person in people
+    ]
+
+    for person, properties in zip(person_list, people):
+        if ("wife" in properties) and (properties["wife"] is not None):
```
ArsenPidhoretskyi, line 28: simply if conditions: just use `properties.get("wife") is not None`
The same with husband

ArsenPidhoretskyi (changes requested):

iishchenko:
Changing: 
simply if conditions: just use properties.get("wife") is not None
The same with husband

app/main.py
```diff
@@ -1,8 +1,35 @@
...
+        people: list[dict[str, str | int | None]]
+) -> list[Person]:
+    person_list = [
+        Person(person["name"], person["age"])
+        for person in people
+    ]
+
+    for person, properties in zip(person_list, people):
+        if properties.get("wife") is not None:
+            person.add_wife(properties["wife"])
```
ArsenPidhoretskyi, line 31: Really, you do not need a method for this action. Because you have the same checking in this method(for None). Better to do in this line: `person.wife = self.people[properties["wife"]]` and you can remove method `add_wife`
The same with husband

ArsenPidhoretskyi: As for me, it the most simple way

ArsenPidhoretskyi (changes requested):

iishchenko:
changed:
Really, you do not need a method for this action. Because you have the same checking in this function(for None). Better to do in this line: person.wife = self.people[properties["wife"]] and you can remove method add_wife
The same with husband

ArsenPidhoretskyi (approved):
Super! Well done

PR #341, opened 2023-01-23
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+
+    people = {}
```
ArsenPidhoretskyi, line 3: Add type annotations here

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(human["name"], human["age"]) for human in people]
+    for human in people:
+        if "wife" in human and human["wife"] is not None:
```
ArsenPidhoretskyi, line 14: You can simplify condition by using .get method on the dict
The same with husband

ArsenPidhoretskyi (changes requested):

PR #344, opened 2023-01-24
Solution

IgoraStar:


app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
```
ArsenPidhoretskyi, line 2: provide type annotations here

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
ArsenPidhoretskyi, line 7: You do not need to do with self.__class__.people
Two options:
1. self.people
2. Person.people

app/main.py
```diff
@@ -1,8 +1,20 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_list = []
```
ArsenPidhoretskyi, line 11: Update naming. Use people_object or person_list

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    for person in people:
+        new_list.append(Person(name=person["name"], age=person["age"]))
+    for index, person in enumerate(people):
+        if "wife" in person and person["wife"] is not None:
```
ArsenPidhoretskyi, line 15: You can use .get method in dict and simplify the condition
The same with husband

ArsenPidhoretskyi (changes requested):

app/main.py
Bohdan-Salamakha: I see, you haven't resolved this issue yet, but asked for review.
**Please don't hesitate to write to mentor if u have some questions.**
btw Arsen means smth like that
`class Person:`
`    people: dict = {}`

Bohdan-Salamakha (changes requested):
v2

app/main.py
IgoraStar: 1. self.people is an object class but not object attribute
2. this is bad hardcoding practice

Vasyl-Smutok (approved):
Good Job 👍

Bohdan-Salamakha (approved):

PR #346, opened 2023-01-26
"Solution"

numerodiciannove:


app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
```
katevasilenko, line 2: Please provide type annotation here

app/main.py
```diff
@@ -1,8 +1,21 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = []
+    for person in people:
+        person = Person(name=person["name"], age=person["age"])
+        people_list.append(person)
+
+    for i, person in enumerate(people):
+        if "wife" in person.keys() and person["wife"] is not None:
```
katevasilenko, line 17: `.keys()` is unnecessary. Python use it by default

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+            people_list[i].wife = Person.people.get(person["wife"])
+        elif "husband" in person.keys() and person["husband"] is not None:
```
katevasilenko, line 19: Please remove `.keys()`

katevasilenko (changes requested):

app/main.py
numerodiciannove: Type annotation for what? For class attribute people? Why are there often no examples in the lessons and video theory...((( Thank you! I didn't know about type annotation for class atributes.

Abnormaltype (approved):
Well done!

PR #347, opened 2023-01-27
Solution

Thirteenthskyi:


app/main.py
```diff
@@ -1,8 +1,18 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(name=i["name"], age=i["age"]) for i in people]
+    Person.people = {exemplar.name: exemplar for exemplar in person_list}
```
YuriiKindrat, line 11: make more informative name for variable "i"
u can add person to the people dict inside init method - no need to make +1 loop for this

app/main.py
```diff
@@ -1,8 +1,18 @@
...
+
+    for man in people:
```
YuriiKindrat, line 13: maybe it is better to name variable person instead of man ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people.update({self.name: self})
```
ArsenPidhoretskyi, line 7: Why you need an update when you can do:
1. `Person.people[self.name] = self`
2. `self.people[self.name] = self`
The result will be the same

app/main.py
```diff
@@ -1,8 +1,23 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(
+        name=person["name"],
+        age=person["age"]
+    ) for person in people]
```
ArsenPidhoretskyi, line 14:
```suggestion
    person_list = [
        Person(
            name=person["name"],
            age=person["age"]
        ) 
        for person in people
    ]
```

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+
+    for person in people:
+        if person.get("wife") is not None:
+            Person.people[person["name"]].wife = \
+                Person.people[person["wife"]]
```
ArsenPidhoretskyi, line 19: Write inline or move Person.people[person["wife"]] to variable and then set
The same with husband

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,24 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [
+        Person(
+            name=person["name"],
+            age=person["age"]
+        )
+        for person in people]
```
YuriiKindrat, line 16: fix identation here, list comprehesion closing bracket should be at the new line

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #348, opened 2023-01-28
'Solution'

natalie-goriela:


app/main.py
```diff
@@ -1,8 +1,34 @@
...
+            if person.name == \
+                    one_person.get("name") and one_person.get("wife"):
+                person.wife = \
+                    find_partner(one_person.get("wife"), persons)
+            if person.name == \
+                    one_person.get("name") and one_person.get("husband"):
+                person.husband = \
+                    find_partner(one_person.get("husband"), persons)
+
+    return persons
```
Serhii-Leonenko, line 34: You have class attribute people, so you don't need a redundant loop for searching for a partner, just take it from the dict with the name key

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,27 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    persons = []
+
+    for one_person in people:
+        persons.append(Person(one_person["name"], one_person["age"]))
```
ArsenPidhoretskyi, line 16: Use list comprehensions. Also, one_person is bad naming, because person is already one

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    persons = []
```
ArsenPidhoretskyi, line 13: 'persons' is single people is plural

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+
+    for one_person in people:
+        persons.append(Person(one_person["name"], one_person["age"]))
+
+    for person in persons:
+        for one_person in people:
```
ArsenPidhoretskyi, line 19: Double loop is not a good idea

ArsenPidhoretskyi: You save all people in Person.people

ArsenPidhoretskyi (changes requested):

andrewbelo (changes requested):
Все-таки, перейменуй `one_person` в просто `person` це і є однина від слова піпл

app/main.py
```diff
@@ -1,8 +1,25 @@
...
-    pass
+
+    person_list = [
+        Person(person["name"], person["age"]) for person in people
+    ]
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = \
+                Person.people[person["wife"]]
```
olena-hul, line 20: use variables to avoid splitting it into multiple lines. Fix for husband too

olena-hul (changes requested):

stanhjr (approved):

PR #352, opened 2023-01-31
Solution

WalDragnir:


Abnormaltype (approved):
Well done!

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+def create_person_list(persons: list) -> list:
+    person_array = []
+    for person in persons:
+        person_array.append(Person(name=person["name"], age=person["age"]))
+    for person in persons:
+        if "wife" in person.keys() and person["wife"] is not None:
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        elif "husband" in person.keys() and person["husband"] is not None:
+            Person.people[person["name"]].husband \
+                = Person.people[person["husband"]]
```
YuriiKindrat, line 19: u can create a separate variable with Person.people[person["name"]] to write less code,so u can remove line breaking from 18 line

app/main.py
```diff
@@ -1,8 +1,20 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(persons: list) -> list:
+    person_array = []
+    for person in persons:
+        person_array.append(Person(name=person["name"], age=person["age"]))
+    for person in persons:
+        if "wife" in person.keys() and person["wife"] is not None:
```
YuriiKindrat, line 15: also u can use method get here, to avoid using and here, the same is for husband

YuriiKindrat (changes requested):

app/main.py
Abnormaltype, line 15: `keys() ` is extra here

Abnormaltype (changes requested):

vasylhnatiuk (approved):
good job!

PR #357, opened 2023-02-03
1 commit

AndereLion:


Maytreyaya (approved):
Good job

app/main.py
```diff
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: str) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
wQuelS, line 7: You should access class attribute through class name not `self`

wQuelS (changes requested):

app/main.py
AndereLion: Fixed

YuriiKindrat (approved):

PR #358, opened 2023-02-04
Solution

SafonovVladimir:
add my Solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -36,7 +36,7 @@ def add_husband(people: list, list_person: list) -> None:
             for wife in list_person:
                 for husband in list_person:
                     if (
-                            person["husband"] == husband.name
-                            and person["name"] == wife.name
+                        person["husband"] == husband.name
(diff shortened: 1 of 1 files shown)


DavidGlusch (approved):

app/main.py
```diff
@@ -1,8 +1,28 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
wQuelS, line 7: You should access class attribute through class name not `self`

wQuelS (changes requested):

app/main.py
SafonovVladimir: Fixed it

Zenionoi (approved):

YuriiKindrat (approved):

PR #359, opened 2023-02-04
Solution

bihunva:


app/main.py
```diff
@@ -1,8 +1,34 @@
...
+        if check_marital_status(person) == "married":
+            if check_gender(person) == "man":
+                instance.wife = Person.people[person["wife"]]
+            else:
+                instance.husband = Person.people[person["husband"]]
+
+    return list_of_instances
+
+
+def check_gender(person: dict) -> str:
```
AndereLion, line 25: U can use only one return here. It will be better approach. =)  
if "wife" in person:
        return "man"
    return "woman"

app/main.py
```diff
@@ -1,8 +1,34 @@
...
+    if "wife" in person:
+        return "man"
+    return "woman"
+
+
+def check_marital_status(person: dict) -> str:
+    if person.get("husband") is not None or person.get("wife") is not None:
```
AndereLion, line 32: The same.

AndereLion (changes requested):

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+
+    return list_of_instances
+
+
+def check_gender(person: dict) -> str:
+    return "man" if "wife" in person else "woman"
+
+
+def check_marital_status(person: dict) -> str:
+    return "married" if person.get("husband") is not None \
```
AndereLion, line 30: We are not allowed to move the string like this. With using " \ ". Must be like this (
code
code
)
or like this ( code
code)

AndereLion (changes requested):

AndereLion (approved):
Good job !

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_instances = [Person(p["name"], p["age"]) for p in people]
```
wQuelS, line 11: `p` is not descriptive name

wQuelS (changes requested):

wQuelS (approved):
Well done!

PR #360, opened 2023-02-04
Solution

ZabFTFT:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_instances = []
```
manjustice, line 12: There is no need to create a variable here

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        people_instances.append(person_instance)
+
+    for person in people:
+        class_people_person_name = Person.people[person["name"]]
+        if "wife" in person and person["wife"] is not None:
+            class_people_person_name.wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"] is not None:
+            class_people_person_name.husband = Person.people[person["husband"]]
+
+    return people_instances
```
manjustice, line 25: You can return attribute people of Person's class

manjustice (changes requested):

app/main.py
ZabFTFT: But how? We have a dict as class attribute not a list -> "This class also should have a class attribute people, it is a dict that stores Person instances by their name"

manjustice: return list(Person.people.values())

ZabFTFT: I see, thank you!

manjustice (approved):
Well done!

app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
```
wQuelS, line 8: You should access class attribute through class name (`Person`) not self

app/main.py
```diff
@@ -1,8 +1,23 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    for person in people:
+        Person(person["name"], person["age"])
+
+    for person in people:
+        class_people_person_name = Person.people[person["name"]]
+        if "wife" in person and person["wife"] is not None:
```
wQuelS, line 18: Can you simplify this by using get() method?

wQuelS (changes requested):

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #362, opened 2023-02-05
Solution

DavidGlusch:


ZabFTFT (approved):
Looks okay for me. Mb its better to separate with return statement and body of function but overall "Good job!"

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for index, person in enumerate(people):
+        result.append(Person(person["name"], person["age"]))
+        if "wife" in person and person["wife"] in Person.people:
```
wQuelS, line 14: Can you simplify this check? Maybe get() can help you ?

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+            result[index].wife = Person.people[person["wife"]]
+            Person.people[person["wife"]].husband = result[index]
+        elif "husband" in person and person["husband"] in Person.people:
```

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #365, opened 2023-02-06
'Solution'

alexlega:


app/main.py
```diff
@@ -1,8 +1,30 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people.setdefault(self.name, self)
```
Serhii-Leonenko, line 7: You don't need setdefault method here, because you don't need to return value if it already exist

app/main.py
```diff
@@ -1,8 +1,30 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_instances = []
+    for i in range(len(people)):
+        list_instances.append(Person(people[i]["name"], people[i]["age"]))
+    for first_person in range(len(people)):
+        if "wife" in people[first_person].keys():
```
Serhii-Leonenko, line 18: Use just `dict.get()` method in the condition

Serhii-Leonenko (changes requested):

Y-Havryliv (approved):
Well done!

PR #366, opened 2023-02-06
solution

usrofgh:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_objects = [Person(person["name"], person["age"])
+                      for person in people]
+
+    for person in people:
+        if "wife" in person and person["wife"]:
```
Serhii-Leonenko, line 15: Try to use only `dict.get()` method in the condition.

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+            Person.people[person["name"]].wife = \
+                Person.people[person["wife"]]
```
Serhii-Leonenko, line 17: Don’t use `= \` , create temporal variable instead

app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
Serhii-Leonenko, line 7: You should access class attribute through class name not self

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
wQuelS, line 7: You should access class attribute through class name not `self`.
For this case it is `Person...`

app/main.py
```diff
@@ -1,8 +1,23 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_objects = [Person(person["name"], person["age"])
+                      for person in people]
+
+    for person_el in people:
```
wQuelS, line 14: why not just `person` ?

wQuelS (changes requested):

app/main.py
usrofgh: oops.. Forgot to rename

app/main.py
usrofgh: So, do I understand correctly that in ```__init__``` we always use ClassName.attribute ? We do never use this kind of construction ```self.__class__.attribute``` ?

Y-Havryliv (approved):
Well done!

PR #367, opened 2023-02-06
Solution

WDemidenko:


NazarHladaniuk (approved):

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        person_list.append(Person(person["name"], person["age"]))
+    for i in range(len(people)):
```
wQuelS, line 14: "i" is very bad variable name. I know you wanted to iterate with index, but anyway.

P.S. you don't really need `for i in range..` loop

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        person_list.append(Person(person["name"], person["age"]))
```
wQuelS, line 13: Can you make it with list comprehension?

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+    for i in range(len(people)):
+        if people[i].get("wife") and people[i]["wife"] is not None:
```
wQuelS, line 15: This check is tooo overcomplicated. I bet you can quickly come up with an easier solution if you only remember what returns `get()` method if there is no such key in dict

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+            person_list[i].wife = Person.people[people[i]["wife"]]
+        if people[i].get("husband") and people[i]["husband"] is not None:
```

wQuelS (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #368, opened 2023-02-06
Solution

SergiiMachulin:


app/main.py
```diff
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int, **kwargs) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
kuznetsb, line 7: As it studied at Mate lectures, It's better to use just Person.people to add element to class attribute instead of using magic method __class__

kuznetsb (changes requested):

app/main.py
SergiiMachulin: OK

kuznetsb (approved):

kuznetsb (approved):
Good job!

YuriiKindrat (approved):

PR #369, opened 2023-02-06
Solution

kuznetsb:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = [Person(name=human["name"], age=human["age"])
+               for human in people]
+    for human in people:
+        if "wife" in human and human["wife"]:
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
+        elif "husband" in human and human["husband"]:
+            Person.people[human["name"]].husband\
```
DmytryTong, line 17: I'd prefer to create sume variables to contain a long-named data units like "Person.people[human["name"]].wife etc. Flake8 app is not complaining about that, as I see, but it's a good thing for making your code easy to read.

DmytryTong (approved):
There is some style problems (too long naming), so fix that please, but in general - its working and looking nice.

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = [Person(name=human["name"], age=human["age"])
+               for human in people]
+    database = Person.people
```
YuriiKindrat, line 13: no need to create one more variable for Person.people

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+    for human in people:
+        if "wife" in human and human["wife"]:
+            database[human["name"]].wife = database[human["wife"]]
+        elif "husband" in human and human["husband"]:
+            database[human["name"]].husband = database[human["husband"]]
```
YuriiKindrat, line 18: u can use get method of a dict in ur condition, so there won't need to use and keyword

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,19 @@
...
-    # write your code here
-    pass
+    persons = [Person(name=human["name"], age=human["age"])
+               for human in people]
+    for human in people:
+        if human.get("wife"):
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
+        elif human.get("husband"):
+            Person.people[human["name"]].husband = \
+                Person.people[human["husband"]]
```
YuriiKindrat, line 18: u can create variable for this part of code, to avoid line breaking Person.people[human["name"]]

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,19 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = [Person(name=human["name"], age=human["age"])
+               for human in people]
+    for human in people:
+        if human.get("wife"):
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
+        elif human.get("husband"):
+            person_husband = Person.people[human["husband"]]
```
YuriiKindrat, line 17: it is better to make a variable for Person.people[human["name"]] cause u are using it for 2 times in ur code

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #370, opened 2023-02-06
Solution but need help to get rid of almost identical ifs

NEkropo1:


inrock1 (approved):
I don't have any remarks.

there is not needed reduse the number of 'if'.
but if you want less, you can review my solution,

NEkropo1:
> 

Well, I tried to make some inner func, which makes same algorithm for both ending if's, but it was so much overloaded that i skipped it for chance of finding more efficient solution(without double loops)

YuriiKindrat (approved):

PR #371, opened 2023-02-06
solution

amber-marichi:


app/main.py
```diff
@@ -1,8 +1,32 @@
...
+            partner_name = person_info["husband"]
+            partner_status = "husband"
+        if partner_status is not None and partner_name in Person.people:
+            partner_entry = Person.people.get(partner_name)
+            if partner_status == "wife":
+                person_entry.wife = partner_entry
+                partner_entry.husband = person_entry
+            else:
+                person_entry.husband = partner_entry
+                partner_entry.wife = person_entry
```
Morty67, line 28: can you make this code shorter?

Morty67 (changes requested):

Morty67 (approved):

app/main.py
```diff
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+    people = dict()
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
```
wQuelS, line 7: you should get access to class attributes using class name (`Person` in this case), not `self`

app/main.py
```diff
@@ -1,8 +1,24 @@
...
-    # write your code here
-    pass
+    people_list = list()
+
+    for person_info in people:
+        person_entry = Person(name=person_info["name"], age=person_info["age"])
+        people_list.append(person_entry)
+
+    for person in people:
+        if person.get("husband") is not None:
```
wQuelS, line 18: `is not None` redundant here, since in your case `if` statement will be `True` if key exists

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+            Person.people[person["name"]].husband \
```
wQuelS, line 19: try to avoid line breaks. You can break line by " [] "

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+                = Person.people[person["husband"]]
+        elif person.get("wife") is not None:
```
wQuelS, line 21: same with `is not None`

wQuelS (changes requested):

YuriiKindrat (changes requested):

Y-Havryliv (approved):
Nice!

PR #373, opened 2023-02-06
Develop

inrock1:


oksana-feshchenko (approved):
Its working correctly, but not the shortest code

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.age = age
+        self.__class__.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
+    person_list = []
+    for prs in people_list:
```
YuriiKindrat, line 12: prs is a bad name for variable, use full names for naming

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        spouse = prs.get("wife") or prs.get("husband")
+        if spouse:
+            if spouse in Person.people:
+                spouse_link = Person.people[spouse]
+                if "wife" in prs:
+                    person.wife = spouse_link
+                    spouse_link.husband = person
+                else:
+                    person.husband = spouse_link
+                    spouse_link.wife = person
```
YuriiKindrat, line 23: try to change logic of ur code, u are checking for wife in prs for 2 times

YuriiKindrat (changes requested):

app/main.py
inrock1: ok, I changed the name. 
But in this case, the short name was intentional because there was
class = Person,
class attribute = people,
list = people
and with such a name, the variable in the for loop confuses all of this code for me finally.

app/main.py
inrock1: Exactly.
I removed the first condition.

app/main.py
```diff
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
wQuelS, line 7: you should get access to class attribute using `Person` not `self`

app/main.py
```diff
@@ -1,8 +1,24 @@
...
-    pass
+
+def create_person_list(people_list: list) -> list:
+    person_list = []
+    for human in people_list:
+        person = Person(human["name"], human["age"])
+        spouse = human.get("wife") or human.get("husband")
+        if spouse in Person.people:
+            spouse_link = Person.people[spouse]
+            if "wife" in human:
```
wQuelS, line 17: you still doing "double-check" and overall logic is a bit too complicated

wQuelS (changes requested):

app/main.py
inrock1: corrected

app/main.py
inrock1: corrected

Serhii-Leonenko (approved):
Don't make extra PRs.

PR #375, opened 2023-02-07
Solution

Zenionoi:


Danil228 (approved):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_ = list()
+    for person in people:
+        list_.append(Person(person["name"], person["age"]))
+        Person.people[person["name"]] = list_[-1]
+    for i, person in enumerate(people):
```
YuriiKindrat, line 14: i is a bad name for variable use full names for naming, f.e index or person_index

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.age = age
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_ = list()
+    for person in people:
+        list_.append(Person(person["name"], person["age"]))
+        Person.people[person["name"]] = list_[-1]
```
YuriiKindrat, line 13: u can make this in init method, so instance will be added authomaticaly

YuriiKindrat (changes requested):

PR #379, opened 2023-02-07
Solution

MikhailLyvak:


oksanaaam (approved):
Nice job)

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    peoples_lst = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```
wQuelS, line 15: I bet you can quickly come up with easier check statement. Just remember or find out what `get()` method returns if there is no such key in dictionary

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        if "husband" in person and person["husband"] is not None:
```

wQuelS (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    pass
+    peoples_lst = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if "wife" in person and person.get("wife") is not None:
+            Person.people[person.get("name")].wife = \
+                Person.people[person.get("wife")]
+        if "husband" in person and person.get("husband") is not None:
+            Person.people[person.get("name")].husband = \
+                Person.people[person.get("husband")]
```
YuriiKindrat, line 20: no need to use "wife/husband" condition here, u already check this using get method, 
also try not to use line breakers, u can make separate variable for this Person.people[person["name"]] to avoid line breaking, or split line by [] brackets

app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    pass
+    peoples_lst = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if person.get("wife") is not None:
+            name_give_w = Person.people[person.get("wife")]
+            Person.people[person.get("name")].wife = name_give_w
+        if person.get("husband") is not None:
+            name_give_h = Person.people[person.get("husband")]
+            Person.people[person.get("name")].husband = name_give_h
```
Serhii-Leonenko, line 20: use more descriptive names for variables, here you can use `wife` and `husband` instead of these name_give...

Serhii-Leonenko (changes requested):

PR #380, opened 2023-02-07
Solution

yura-hudzovskyi:


oigod (approved):

wQuelS (approved):
kwargs here is a bit of overkill, but solution is good.
Well done!

PR #381, opened 2023-02-07
Solution

TeliksPG:


yura-hudzovskyi (approved):

app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
```
wQuelS, line 8: you should get access to class attribute using `Person` not `self`

app/main.py
```diff
@@ -1,8 +1,21 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_person = [
+        Person(person["name"], person["age"]) for person in people
+    ]
+    for name, person in Person.people.items():
+        for info in people:
```
wQuelS, line 16: loop in loop.. Isn't it too complicated?

wQuelS (changes requested):

app/main.py
ProdyRodion: U may suggest anything else?

app/main.py
ProdyRodion: no problem

wQuelS (approved):
GJ

PR #382, opened 2023-02-07
Solution

oigod:


app/main.py
```diff
@@ -1,8 +1,33 @@
 class Person:
-    # write your code here
-    pass
```
dogmoony, line 3: you need to delete this "pass"

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+        Person.people[self.name] = self
+
+    @classmethod
+    def get_people_by_name(cls, people_name: str) -> object:
+        return cls.people.get(people_name)
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
```
dogmoony, line 8: delete this "pass"

dogmoony (changes requested):

app/main.py
oigod: its already deleted

app/main.py
oigod: same

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+    person_objects = []
+    for raw_person in people:
+        person = Person(raw_person["name"], raw_person["age"])
+        person_objects.append(person)
```
Serhii-Leonenko, line 22: Try to use list comprehension for that purpose

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+
+    for raw_person in people:
+        person = Person.people.get(raw_person["name"])
```
Serhii-Leonenko, line 25: You have implemented classmethod, but don't use it here, why?)

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,31 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_objects = [Person(raw_person["name"], raw_person["age"])
+                      for raw_person in people]
+
+    for raw_person in people:
+        person = Person.get_people_by_name(raw_person["name"])
+        if not person:
+            continue
```
YuriiKindrat, line 25: no need to make this condition here

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #383, opened 2023-02-07
Solution

ykopatko:


app/main.py
```diff
@@ -1,8 +1,19 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
oksana0410, line 7: Perhaps Person.people[self.name] = self will be better

oksana0410 (approved):

app/main.py
wQuelS: yep

wQuelS (changes requested):

wQuelS (approved):
Nice work

PR #384, opened 2023-02-07
'Solution'

oksanaaam:


app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = [Person(person["name"], person["age"]) for person in people]
+    for person in people:
+        wife = person.get("wife")
```
AlessaQ, line 13: Line can be inserted directly to condition's. 
Also same in "husband"

AlessaQ (changes requested):

app/main.py
oksanaaam: yes, but than will be bad reading code and overcharacters in line

PR #387, opened 2023-02-07
solution. 

Yulia-Mala:
It is not precise annotation when I return object in pick_by_name method, but if I use from future import annotations - I have 2 tests failed and strange errors with "ast.parse". For now, it's too much, so I've cheated. sorry.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
+
+    @classmethod
+    def pick_by_name(cls, in_name: str) -> object:
+        for person in Person.people:
+            if person == in_name:
+                return Person.people[person]
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    instances_ls = [Person(person["name"], person["age"]) for person in people]
+    for person in people:
+        pair = person.get("wife") or person.get("husband")
+        name = person["name"]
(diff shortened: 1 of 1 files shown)


viktoria-rybenchuk (approved):

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
+
+    @classmethod
+    def pick_by_name(cls, in_name: str) -> object:
+        for person in Person.people:
+            if person == in_name:
+                return Person.people[person]
```
Serhii-Leonenko, line 14: You don't need to use a loop here, because you spend extra time for it every time you call this method. You can get the value from the dict with the name as a key.

Serhii-Leonenko (changes requested):

app/main.py
Yulia-Mala: yes)))) thanks))

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    instances_ls = [Person(person["name"], person["age"]) for person in people]
+    for person in people:
+        pair = person.get("wife") or person.get("husband")
```
YuriiKindrat, line 14: no need to create variable for this, u can make this in if condition, so there won't need to check "wife/husband" in person

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #392, opened 2023-02-07
'Solution'

NickSavchyn:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_list = []
+
+    for human in people:
+        new_human = Person(human["name"], human["age"])
+        new_list.append(new_human)
+
```
skmiwa1, line 18: мені нравицця!

skmiwa1 (approved):

app/main.py
Nattalli: @skmiwa1 not required, but it's better when you make comments in English (for English improvement and practice + maybe on the job you will need to communicate in English)

app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_list = []
+
+    for human in people:
+        new_human = Person(human["name"], human["age"])
+        new_list.append(new_human)
```
Nattalli, line 17: can you make it with list comprehension?

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife =\
+                Person.people[person["wife"]]
```
Nattalli, line 22: I don't recommend you use `\` for writing in a new line. For example, you can make `Person.people[person["name"]]` as a variable

Nattalli (changes requested):

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+    new_list = [Person(human["name"], human["age"]) for human in people]
+
+    for person in people:
+        person_name = Person.people[person["name"]]
+        if person.get("wife"):
+            person_name.wife =\
+                Person.people[person["wife"]]
+        if person.get("husband"):
+            person_name.husband =\
+                Person.people[person["husband"]]
```
wQuelS, line 23: u still using line breaks, which i recommend to avoid. Can you use, maybe, additional variables or some other approaches to get rid of those line breaks ?

wQuelS (changes requested):

PR #397, opened 2023-02-07
Solution

me1nyk:


NickSavchyn (approved):

app/main.py
```diff
@@ -1,8 +1,18 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons_list = [Person(person["name"], person["age"]) for person in people]
+    for name, person in Person.people.items():
+        for personal_info in people:
+            if personal_info.get("wife") and name == personal_info["name"]:
+                person.wife = Person.people[personal_info["wife"]]
+            if personal_info.get("husband") and name == personal_info["name"]:
+                person.husband = Person.people[personal_info["husband"]]
```
wQuelS, line 17: you can simplify this solution. No need in nested loops

wQuelS (changes requested):

app/main.py
me1nyk: thanks for remark, now its easier to read, i believe...
can you also advice please, is it a "bad practice" to make empty string before both "if" ?
anyway, string have some space because of loop, but its gonna be more readable on my mind

wQuelS: No it is not a bad practice. It depends on specification of the project you working on. In this case - it is up to you

app/main.py
```diff
@@ -1,8 +1,19 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons_list = [Person(person["name"], person["age"]) for person in people]
+    for personal_info in people:
+        if personal_info.get("wife"):
+            Person.people[personal_info["name"]].wife = \
+                Person.people[personal_info["wife"]]
```
wQuelS, line 15: get rid of those line breaks. You have almost always avoid using them. You can use additional variables to do so

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        if personal_info.get("husband"):
+            Person.people[personal_info["name"]].husband = \
+                Person.people[personal_info["husband"]]
```
wQuelS, line 18: same

wQuelS (changes requested):

wQuelS (approved):
Really nice !

PR #399, opened 2023-02-07
Solution

tytomyr:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.name = name
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for one in people:
```
skmiwa1, line 12: it doesn's seem propriate name to me (one)

skmiwa1 (approved):

app/main.py
```diff
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> dict:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
Serhii-Leonenko, line 7: You should access class attribute through class name not self

app/main.py
Serhii-Leonenko: yes, fix the name

app/main.py
```diff
@@ -1,8 +1,22 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for one in people:
+        Person(one["name"], one["age"])
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```
Serhii-Leonenko, line 15: Use only `dict.get()` method in the condition

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+            Person.people[person["name"]].__dict__["wife"] \
+                = Person.people[person["wife"]]
```
Serhii-Leonenko, line 17: Don’t use `= \` , create temporal variable instead

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person in people:
+        Person(person["name"], person["age"])
+    for person in people:
+        individual = Person.people[person["name"]]
+        if person.get("wife") and person["wife"] is not None:
```
Nattalli, line 17: Look:
```python
if person.get("wife") and person["wife"] is not None:
```
is the same:
```python
if person.get("wife"):
```
Read a theory about that. And change this condition (as the second one is unnecessary)

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+            individual.__dict__["wife"] = Person.people[person["wife"]]
+        if person.get("husband") and person["husband"] is not None:
```
Nattalli, line 19: and the same here

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person in people:
+        Person(person["name"], person["age"])
```
Nattalli, line 14: can you make it with list comprehension?

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+    for person in people:
+        individual = Person.people[person["name"]]
+        if person.get("wife") and person["wife"] is not None:
+            individual.__dict__["wife"] = Person.people[person["wife"]]
+        if person.get("husband") and person["husband"] is not None:
+            individual.__dict__["husband"] = Person.people[person["husband"]]
+        person_list.append(Person.people[person["name"]])
```
Nattalli, line 21: why do you do this append here? It is more logical to make in the first loop `$13-14` lines

Nattalli (changes requested):

Nattalli (approved):
well done! 👍🏻

PR #404, opened 2023-02-08
Solution

destr34:


app/main.py
```diff
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
```
Yevheniy-Lototskiy, line 8: You can just write 'name' here without 'self.'

Yevheniy-Lototskiy (approved):

app/main.py
```diff
@@ -1,8 +1,24 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_persons = []
+    for person_for_dict in people:
+        person = Person(person_for_dict["name"], person_for_dict["age"])
+        list_of_persons.append(person)
```
Nattalli, line 15: make it with list comprehension

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        if person.get("husband"):
+            Person.people[person["name"]].husband = \
+                Person.people[person["husband"]]
```
Nattalli, line 22: Don’t use `= \` , create temporal variable instead

Nattalli (changes requested):

app/main.py
destr34: at the first leson it was not so obviouse for me)

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_persons = [(Person(_["name"], _["age"])) for _ in people]
+
+    for _ in people:
```
YuriiKindrat, line 14: it is bad practice to use "_" variable for loop if u will use it later, use "_" only if there is no need to use this variable

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #405, opened 2023-02-08
Solution

VladyslavCherkashyn:


skmiwa1 (approved):
as I mentioned before solution without nested for loops exists//lets see what will mentors say

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+                        people[crash]["name"]:
+                    links_of_instances[waifu].wife = \
+                        links_of_instances[crash]
+        if "husband" in people[waifu].keys():
+            for crash in range(len(people)):
+                if people[waifu]["husband"] == \
+                        people[crash]["name"]:
+                    links_of_instances[waifu].husband = \
+                        links_of_instances[crash]
+
```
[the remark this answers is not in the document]
VladyslavCherkashyn: Unfortunately, I didn`t find another way to get the necessary values ​​to the keys  "wife" and "husband".
Considered options through the get(value) but the loop remains with the same nesting levels.
If you have an algorithm on how to make it easier - please advise.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        self.age = age
+        Person.people.setdefault(self.name, self)
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    links_of_instances = [
+        Person(people[person]["name"], people[person]["age"])
+        for person in range(len(people))]
```
YuriiKindrat, line 13: fix indentation here, also no need to use range/len methods here

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+                if people[waifu]["wife"] == \
+                        people[crash]["name"]:
+                    links_of_instances[waifu].wife = \
+                        links_of_instances[crash]
+        if "husband" in people[waifu].keys():
+            for crash in range(len(people)):
+                if people[waifu]["husband"] == \
+                        people[crash]["name"]:
+                    links_of_instances[waifu].husband = \
+                        links_of_instances[crash]
```
YuriiKindrat, line 27: try to simplify ur logic - no need to use range/len methods here, no need to use for loop for 2 times, no need to use keys method, it is used by default, try to use ur people - class attribute

YuriiKindrat (changes requested):

Y-Havryliv (approved):
Nice!

PR #407, opened 2023-02-08
'Solution'

Bohdan-Kovalenko1992:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = []
+    for person in people:
+        person = Person(name=person["name"], age=person["age"])
+        people_list.append(person)
```
oksanaaam, line 15: you can write these 4 lines in one: people_list = [the same code ... in people]

oksanaaam (changes requested):

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = [Person(person["name"], person["age"]) for person in people]
+
+    for i, person in enumerate(people):
```
wQuelS, line 14: `i` is bad variable name

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        if "wife" in person and person["wife"] is not None:
```
wQuelS, line 15: you can use dictionary method `get`. It is safe method, 'cause in case if there is no such key in dictionary it will return `None`. That's exactly what u need, I guess

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+            people_list[i].wife = Person.people.get(person["wife"])
+        if "husband" in person and person["husband"] is not None:
```

wQuelS (changes requested):

app/main.py
wQuelS: emm.. what about `get()` method ?

wQuelS (changes requested):

wQuelS (approved):
Here we go, Good job!

PR #408, opened 2023-02-08
Solution

lotrik84:


app/main.py
```diff
@@ -1,8 +1,19 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
```
vasylhnatiuk, line 7: would be better to use  `Person.people[self.name]`

vasylhnatiuk (approved):

vasylhnatiuk:
Good job!

app/main.py
lotrik84: Thanks, i missed that)

PR #409, opened 2023-02-08
Solution

Litvinenko23:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    instance_list = []
+    for human in people:
```
Yulia-Mala, line 13: you can also use list comprehension here if you want

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        instance = Person(human["name"], human["age"])
+        instance_list.append(instance)
+
+    for instance in instance_list:
+        for human in people:
```
Yulia-Mala, line 18: you dont need nested loop here (I had the same mistake and mentor requested to change it). your list of  dicts has format "name: instance" , and the same name you can pick by human["name"]. so we have: for human in people: instance_list[ human["name"]]... I saved  human["name"]  in additinal var  just for not to mix it up...

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+            if human["name"] == instance.name:
+                if "wife" not in instance.__dict__ and \
```
Yulia-Mala, line 20: what is the purpose of this check ? you dont have another chance in your code to add wife/husband in the __dict__ so you probably can delete it?

Yulia-Mala (changes requested):

Yulia-Mala (approved):
:)

app/main.py
```diff
@@ -1,8 +1,21 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    instance_list = [Person(human["name"], human["age"]) for human in people]
+
+    for human in people:
+        human_name = human["name"]
+        if human.get("wife") is not None:
```
Nattalli, line 16: you can make it easier:
`if human.get("wife") is not None:` <=> `if human.get("wife"):`
that is the same but looks better

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+            Person.people[human_name].wife = Person.people[human["wife"]]
+        if human.get("husband") is not None:
```
Nattalli, line 18: the same here

Nattalli (changes requested):

app/main.py
Serhii-Leonenko: still not fixed. The same about "husband" key

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #415, opened 2023-02-09
Solution

StreetNik:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+
+    for person in people:
+        instance = Person(person["name"], person["age"])
+        for person_info in person:
+            if person[person_info] is not None \
+                    and (person_info == "wife" or person_info == "husband"):
```
Vasyl-Smutok, line 18: better to use `()` then `\`
```suggestion
            if (
                person[person_info] is not None 
                and (person_info == "wife" or person_info == "husband")
            ):
```

Vasyl-Smutok (changes requested):

vasylhnatiuk (approved):
Good job!

PR #417, opened 2023-02-10
solution

tarasavage:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list) -> list[Person]:
+    married_people = []
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+        if "wife" in person.keys() and person["wife"]:
```
Grayder0152, line 14: Instead of this condition you can use `person.get('wife') is not None`

Grayder0152: Apart from this, you can use `:=` operator, but it is not necessary

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+            wife = Person(person["name"], person["age"])
+            wife.husband = new_person
+            new_person.wife = wife
+        elif "husband" in person.keys() and person["husband"]:
```
Grayder0152, line 18: The same as above

app/main.py
```diff
@@ -1,8 +1,23 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list) -> list[Person]:
+    married_people = []
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+        if "wife" in person.keys() and person["wife"]:
+            wife = Person(person["name"], person["age"])
```
Grayder0152, line 15: You need to get person from `people` not create

Grayder0152 (approved):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+            wife.husband = new_person
+            new_person.wife = wife
+        elif "husband" in person.keys() and person["husband"]:
+            husband = Person(person["name"], person["age"])
```
Grayder0152, line 19: The same as above

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list) -> list[Person]:
```
Bohdan-Salamakha, line 10: add additional typehint for list inner objects

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+    married_people = [
+        Person(person["name"], person["age"])
```
Bohdan-Salamakha, line 12: use get method instead of parenthesis everywhere to avoid KeyError

Bohdan-Salamakha (changes requested):
v1

PR #424, opened 2023-02-18
Solution

MkrtychKhachatrian:


app/main.py
```diff
@@ -1,8 +1,20 @@
...
-    pass
+    people = dict()
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
```
Bohdan-Salamakha, line 10: add additional typehint for list inner objects

app/main.py
```diff
@@ -1,8 +1,20 @@
...
-    # write your code here
-    pass
+    people_list = []
+    for person in people:
+        people_list.append(Person(person["name"], person["age"]))
```
Bohdan-Salamakha, line 13: use get() method instead of parenthesis everywhere to avoid KeyError

Bohdan-Salamakha (changes requested):
v1

app/main.py
khachatrian2000: problem
Here I am using get() everywhere I can, but still getting KeyError. Could you, please, give any suggestions?

app/main.py
ArsenPidhoretskyi, line 10: accept list of what
and return list of what?
you can use list[and some type here]

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = []
```
ArsenPidhoretskyi, line 11: people is plural
also use list comprenhension

app/main.py
ArsenPidhoretskyi: @Bohdan-Salamakha No need .get here

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    for person in people:
+        people_list.append(Person(person["name"], person["age"]))
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```
ArsenPidhoretskyi, line 15: You can simplify condition
use `person.get("wife") is not None`
The same with husband

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"] is not None:
+            Person.people[person["name"]].husband = \
+                Person.people[person["husband"]]
```
ArsenPidhoretskyi, line 19: Write inline or do as:
```suggestion
            husband = Person.people[person["husband"]]
            Person.people[person["name"]].husband = husband
```

ArsenPidhoretskyi (changes requested):

khachatrian2000:
> 



> description

Here's solution v2. It works fine, but once I add type annotations for create_person_list() -> I get KeyError

app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    pass
+    people = dict()
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
```
Vasyl-Smutok, line 10:
```suggestion
def create_person_list(people: list[dict]) -> list:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    people_list = [
+        Person(person.get("name"), person.get("age"))
+        for person in people
+    ]
+    for person in people:
+        if person.get("wife") is not None:
+            wife = Person.people[person.get("wife")]
+            Person.people[person.get("name")].wife = wife
```
ArsenPidhoretskyi, line 18: In this case .get inside is redundant

ArsenPidhoretskyi (approved):
Good job!

PR #430, opened 2023-02-27
Develop

MaksNochvai:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
```
Bohdan-Salamakha, line 10: add additional typehint for list inner objects like list[ANY_TYPE]

app/main.py
```diff
@@ -1,8 +1,21 @@
...
-    # write your code here
-    pass
+    people_list = []
+    for person in people:
+        people_list.append(Person(person["name"], person["age"]))
```
Bohdan-Salamakha, line 13: use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError
`person.get("name")`

Bohdan-Salamakha (changes requested):
v1

PR #439, opened 2023-02-28
Solution

antonpiddubnyi:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    pass
+
+def create_person_list(people_initial: list) -> list:
+    people_list = [
+        Person(person["name"], person["age"])
+        for person in people_initial
+    ]
+    for person in people_initial:
+        if "wife" in person and person["wife"] is not None:
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
```
Grayder0152, line 17:
```suggestion
        if wife_name := person.get("wife"):
            Person.people[person["name"]].wife = Person.people[wife_name]
```
Cool approach but it is not necessary

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        elif "husband" in person and person["husband"] is not None:
+            Person.people[person["name"]].husband = \
+                Person.people[person["husband"]]
```
Grayder0152, line 20: The same as above. Do not use `\`, use `()`

Grayder0152 (approved):

PR #447, opened 2023-03-03
'Solution'

kirzenko:


app/main.py
```diff
@@ -1,8 +1,35 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person in people:
+        personage = Person(person["name"], person["age"])\
+            if person["name"] not in Person.people\
+            else Person.people[person["name"]]
```
Grayder0152, line 16: instead of `\` use `()`

Grayder0152: Or do it not in one line

app/main.py
```diff
@@ -1,8 +1,35 @@
...
+        person_list.append(personage)
+
+    for person in people:
+        wife_or_husband = "wife" if "wife" in person else "husband"
+        name_of_wife_or_husband = person[wife_or_husband]
+
+        if name_of_wife_or_husband:
+            wife_husband_object = None
+            if name_of_wife_or_husband not in Person.people:
+                wife_husband_object = Person(name_of_wife_or_husband, 0)
```
Grayder0152, line 26: If `name` not found do nothing

Grayder0152 (changes requested):

PR #449, opened 2023-03-03
Person class

kirzenko:


app/main.py
```diff
@@ -1,8 +1,33 @@
...
+    person_list = [
+        Person.people.get(
+            person.get("name"),
+            Person(person.get("name"), person.get("age"))
+        ) for person in people
+    ]
+
+    for person in people:
+        wife_or_husband = "wife" if "wife" in person else "husband"
+        name_of_wife_or_husband = person[wife_or_husband]
```
ArsenPidhoretskyi, line 20: Not true
Maybe key not exist in total?

ArsenPidhoretskyi: Simply this code. It is too hard
No need to check wife_or_husband
You have only two options:
1. If wife exist 
2. If husband exist

ArsenPidhoretskyi (changes requested):

PR #452, opened 2023-03-06
New solution

kirzenko:


app/main.py
```diff
@@ -1,8 +1,35 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list:
+    person_list = [
+        Person.people.get(
+            person.get("name"),
+            Person(person.get("name"), person.get("age"))
+        ) for person in people
```
ArsenPidhoretskyi, line 15: Why it is so complicated...
```suggestion
        Person(person.get("name"), person.get("age"))
        for person in people
```

app/main.py
```diff
@@ -1,8 +1,35 @@
...
+    ]
+
+    for person in people:
+        if "wife" in person:
+            name_of_wife_or_husband = person["wife"]
+        else:
+            name_of_wife_or_husband = person["husband"]
```
ArsenPidhoretskyi, line 22: Remove this and simplify by just two conditions:
1. if person.get("wife") is not None
2. if person.get("husband") is not None

app/main.py
```diff
@@ -1,8 +1,35 @@
...
+        if name_of_wife_or_husband:
+            if name_of_wife_or_husband not in Person.people:
+                wife_husband_object = Person(name_of_wife_or_husband, 0)
+            else:
+                wife_husband_object = Person.people[name_of_wife_or_husband]
+
+            if "wife":
+                Person.people[person["name"]].wife = wife_husband_object
+            if "husband":
+                Person.people[person["name"]].husband = wife_husband_object
```
ArsenPidhoretskyi, line 33: Btw it is always true

ArsenPidhoretskyi (changes requested):

PR #458, opened 2023-03-09
'Solution'

struhanets:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_per = [Person(body["name"], body["age"]) for body in people]
+    for human in people:
+        if human.get("wife") is not None:
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
+        elif human.get("husband") is not None:
+            Person.people[human["name"]].husband \
```
Grayder0152, line 16: Use `()` instead of `\`

Grayder0152 (approved):

app/main.py
struhanets: I can't do it, I tried but I got an error

PR #459, opened 2023-03-10
Develop

sbilikepy:


app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        self.name = name
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    to_return = []
+    for describe in people:
```
ArsenPidhoretskyi, line 12: descripe - описати?
Use person

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        current_person = Person(describe["name"], describe["age"])
+
+        if "wife" in describe and describe["wife"] in Person.people:
+            create_marriage_by_wife(current_person, describe["wife"])
+
+        if "husband" in describe and describe["husband"] in Person.people:
+            create_marriage_by_husband(current_person, describe["husband"])
```
ArsenPidhoretskyi, line 19: Not correct because describe["husband/wife"] will equal to str. But you must set not name, you must set instance of another Person

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    to_return = []
```
ArsenPidhoretskyi, line 11: "to_return" please choose more correct name

ArsenPidhoretskyi (changes requested):

app/main.py
sbilikepy: Okay, explain me please why we should write code in this way?
"it is a dict that stores Person instances by their name"

vasylhnatiuk (approved):
Good job!

app/main.py
sbilikepy: @ArsenPidhoretskyi

PR #468, opened 2023-03-15
Develop2

dminyaylo:


Vasyl-Smutok (changes requested):
Please fix flake8 error 
image

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+                           husband=husband)
+            husband.wife = human
+        elif "wife" in people_person and people_person["wife"] is not None:
+            wife = Person(people_person["wife"], 0)
+            human = Person(people_person["name"], people_person["age"],
+                           wife=wife)
+            wife.husband = human
+        else:
+            human = Person(people_person["name"], people_person["age"])
+        Person.people[people_person["name"]] = human
```
ArsenPidhoretskyi, line 29: This is not need to do inside create_person list

ArsenPidhoretskyi (changes requested):
Code so hard... You so many times create Person.
Please do it in two loops:
1. Create all person
2. Set to each person husband or wife

app/main.py
```diff
@@ -1,8 +1,33 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int, **kwargs) -> None:
+        self.name = name
+        self.age = age
+        if "husband" in kwargs.keys() and kwargs["husband"] is not None:
+            self.husband = kwargs["husband"]
+        elif "wife" in kwargs.keys() and kwargs["wife"] is not None:
+            self.wife = kwargs["wife"]
```
ArsenPidhoretskyi, line 10: This code is not needed inside init(7-10)

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_people = []
```
ArsenPidhoretskyi, line 15: rename to people_objs

ArsenPidhoretskyi: Also better to use list comprehension

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+
+    for people_person in people:
+        human = Person(people_person["name"], people_person["age"])
+        list_of_people.append(human)
+
+    for people_person in people:
+        human = Person.people[people_person["name"]]
+
+        if "husband" in people_person and people_person["husband"] is not None:
```
ArsenPidhoretskyi, line 24: You can use `if people_person.get `

app/main.py
```diff
@@ -1,8 +1,33 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_people = []
+
+    for people_person in people:
+        human = Person(people_person["name"], people_person["age"])
+        list_of_people.append(human)
+
+    for people_person in people:
```
ArsenPidhoretskyi, line 21: people_person is not properly name
Rename it to person

app/main.py
ArsenPidhoretskyi: The same with wife

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #470, opened 2023-03-16
'Solution'

PetroHavrysh:


app/main.py
```diff
@@ -1,8 +1,25 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: str) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
ArsenPidhoretskyi, line 7: You can simplify this, by using
self.people[name] = self

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+    person_list = []
+    # Створюємо всіх людей, зберігаючи їх у списку person_list
+    for person in people:
+        person_to_list = Person(person["name"], person["age"])
+        person_list.append(person_to_list)
+
+    # Установлюємо партнерів для кожної людини
+    for person in people:
+        person_to_spouses = Person.people[person["name"]]
+        if "wife" in person and person["wife"] is not None:
```
ArsenPidhoretskyi, line 20: Simplify and use person.get

ArsenPidhoretskyi (changes requested):

PR #477, opened 2023-03-22
'Solution'

NataliaStepanenko:


app/main.py
```diff
@@ -1,8 +1,17 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result_list = [Person(person["name"], person["age"]) for person in people]
+    for index, person in enumerate(people):
+        if "wife" in person.keys() and person["wife"] is not None:
```
vasylhnatiuk, line 13: will be better to use `get`:
```suggestion
       if person.get("wife") is not None:
```

app/main.py
```diff
@@ -1,8 +1,17 @@
...
+            result_list[index].wife = Person.people[person["wife"]]
+        if "husband" in person.keys() and person["husband"] is not None:
```
vasylhnatiuk, line 15: the same here

vasylhnatiuk (changes requested):

PR #479, opened 2023-03-25
Solution

un5oul:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person in people:
+        person_list.append(Person(person["name"], person["age"]))
```
olena-hul, line 14: Use llst comprehension

olena-hul (changes requested):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if person.get("wife"):
+            wife = Person.people[person.get("wife")]
+            Person.people[person.get("name")].wife = wife
```
ArsenPidhoretskyi, line 16: .get does not need inside if. Because we know that wife key exist and we know that name is exist also
Fix the same in husband

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        elif person.get("husband"):
```
ArsenPidhoretskyi, line 17: If someone have husband and wife?)

ArsenPidhoretskyi (changes requested):

app/main.py
un5oul: Should I follow the rule "use get method to avoid KeyError" in this case?

un5oul: Should I follow the rule "use get method to avoid KeyError" in this case?

app/main.py
un5oul: mm, just "if" ?

Bohdan-Salamakha (approved):

PR #485, opened 2023-03-31
solution

alim2709:


app/main.py
```diff
@@ -1,8 +1,19 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
ArsenPidhoretskyi, line 7:
```suggestion
        self.people[name] = self
```
It is not needed. You can just use self.people

app/main.py
```diff
@@ -1,8 +1,19 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for person in people:
+        result += {Person(person.get("name"), person.get("age"))}
```
ArsenPidhoretskyi, line 13: Use list comprehension

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+    for i, person in enumerate(people):
```
ArsenPidhoretskyi, line 14: Why you need i

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        if "wife" in person and person.get("wife") is not None:
+            result[i].wife = Person.people.get(person["wife"])
```
ArsenPidhoretskyi, line 16: Objects are mutable. Then if you change it inside
Person.people it will affect result also

app/main.py
```diff
@@ -1,8 +1,19 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for person in people:
+        result += {Person(person.get("name"), person.get("age"))}
+    for i, person in enumerate(people):
+        if "wife" in person and person.get("wife") is not None:
```
ArsenPidhoretskyi, line 15:
```suggestion
        if person.get("wife") is not None:
```
The same with husband

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+    result = [
+        Person(person.get("name"), person.get("age")) for person in people
+    ]
+    i = 0
+    for person in people:
+        if person.get("wife") is not None:
+            result[i].wife = Person.people.get(person["wife"])
+        if person.get("husband") is not None:
+            result[i].husband = Person.people.get(person["husband"])
+        i += 1
```
Bohdan-Salamakha, line 20: use for i in range(len(people)):
to avoid i = 0, i+= 1

u don't need to use C-like counters in Python

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #486, opened 2023-04-03
Solution

vslavchumakov:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
+            wife = Person.people.get(entity["wife"])
+            if wife:
+                person.wife = wife
+                wife.husband = person
+        elif entity.get("husband"):
+            husband = Person.people.get(entity["husband"])
+            if husband:
+                person.husband = husband
+                husband.wife = person
+        person_list.append(person)
```
ArsenPidhoretskyi, line 24: Split it into two loops:
1. Create all entities
2. Set husband/wife inside each people
For now it is unlogical. Because if we have: 
people = [{"name": "a", "age": 25, "wife": "b"}, {"name": "b", "age": 24}]
As a result we will have that a does not have any wife)

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    person_list = [Person(entity["name"], entity["age"]) for entity in people]
+
+    for entity in people:
```
vasylhnatiuk, line 14: person will be a better name
```suggestion
    for person in people:
```

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.name = name
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    person_list = [Person(entity["name"], entity["age"]) for entity in people]
```
vasylhnatiuk, line 12:
```suggestion
    person_list = [Person(person["name"], person["age"]) for person in people]
```

app/main.py
vasylhnatiuk: change it all of the code

vasylhnatiuk (changes requested):

PR #490, opened 2023-04-09
'Solution'

romanfedorchuk:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_people = []
+    for person in people:
+        new_people.append(Person(person.get("name"), person.get("age")))
```
Bohdan-Salamakha, line 13: use list comprehension to do it

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,8 +1,18 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_people = [(Person(person.get("name"), person.get("age")))
+                  for person in people]
+    for i, person in enumerate(people):
+        if "wife" in person and person.get("wife") is not None:
+            new_people[i].wife = Person.people.get(person["wife"])
```
ArsenPidhoretskyi, line 15: No need to do it by index do as follow:
Person.people[person["name"]].wife = Person.people[person["wife"]]

app/main.py
```diff
@@ -1,8 +1,18 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_people = [(Person(person.get("name"), person.get("age")))
+                  for person in people]
+    for i, person in enumerate(people):
+        if "wife" in person and person.get("wife") is not None:
```
ArsenPidhoretskyi, line 14:
```suggestion
        if person.get("wife") is not None:
```
This condition is redundant in this case

app/main.py
```diff
@@ -1,8 +1,18 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_people = [(Person(person.get("name"), person.get("age")))
+                  for person in people]
+    for i, person in enumerate(people):
```
ArsenPidhoretskyi, line 13: You not need enumerate
Check how objects are works. They are mutable or immutable? Do they work by value or link?

ArsenPidhoretskyi (changes requested):
The same errors with husband also

ArsenPidhoretskyi (approved):
Well done!

PR #493, opened 2023-04-09
Create initial implemetation of the module

barbequeue:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for list_person in people:
+        person = Person(list_person.get("name"), list_person.get("age"))
+        person_list.append(person)
+
+    for list_person in people:
```
barbequeue, line 17: Would it be better to complete `wife/husband` assignment in just one iteration through `people` list? Imagine if i had a huge list of people, and then i have to iterate throught the list twice - first time to create all Person, and second - to assign `wife/husband` attributes

olena-hul: You could have tried to do this in 1 iteration)
But it will be a lot of messy code there
You need to assign husband/wife to person and the instance of husband/wife should be already present in the list. That's why we are doing 2 loops here

olena-hul (approved):

olena-hul:
Good job!

PR #496, opened 2023-04-14
Develop

denysbelei9:


app/main.py
```diff
@@ -1,8 +1,20 @@
...
 
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+def create_person_list(people: list[dict[str, str | int | None]]) -> list:
+    person_list = []
+    for human in people:
+        person_list.append(Person(human["name"], human["age"]))
+    for i, human in enumerate(people):
```
ArsenPidhoretskyi, line 15: enumerate is redundant
How objects works? By link or by value?

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        if "wife" in human and human["wife"] is not None:
```
ArsenPidhoretskyi, line 16: Simplify condition and just use human.get

ArsenPidhoretskyi: The same with husband

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+            person_list[i].wife = Person.people.get(human["wife"])
```
ArsenPidhoretskyi, line 17: `Use Person.people[human["name"]].wife = Person.people.get(human["wife"])`

The same with husband

ArsenPidhoretskyi (changes requested):

PR #497, opened 2023-04-15
Develop

denysbelei9:


Bohdan-Salamakha (changes requested):
fix flake8

denysbelei9:
how i can fix flake 8 if string is to longer ,  and if a take half string and push to "a" - it is problem?

PR #503, opened 2023-04-18
Solution

AlyonkaB:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
-    pass
+    list_persons = []
+    for person in people:
+        list_persons.append(Person(
+            name=person["name"],
+            age=person["age"]))
+    for partner in people:
+        if partner.get("wife"):
+            wife = Person.people[partner["wife"]]
+            wife.husband = Person.people[partner["name"]]
```
ArsenPidhoretskyi, line 20: This line is wrong must be:
```suggestion
            Person.people[partner["name"]].wife = wife.husband
```

ArsenPidhoretskyi: The same in husband

app/main.py
```diff
@@ -1,8 +1,24 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_persons = []
+    for person in people:
+        list_persons.append(Person(
+            name=person["name"],
+            age=person["age"]))
```
ArsenPidhoretskyi, line 16: Better to use list comprehension

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: > If person's key wife/husband is not None - create_person_list should add attribute wife/husband respectively to its instance. This attribute should be a link to a Person instance with name the same as wife/husband key in person's dict.

It mush add the key to partner["name"].wife not to the partner["wife"].husband
Please read attentively and check the example

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done!

PR #506, opened 2023-04-23
Solution

vkrasnovyd:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for new_person in people:
+        new_person = Person(new_person["name"], new_person["age"])
```
Bohdan-Salamakha, line 13: u don't need to create a local variable here
```suggestion
    for person in people:
        Person(person["name"], person["age"])
```

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    for person in people:
+        Person(person["name"], person["age"])
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        elif person.get("husband"):
+            Person.people[person["name"]].husband =\
+                Person.people[person["husband"]]
+    return [person for person in Person.people.values()]
```
ArsenPidhoretskyi, line 21: It is not true because for example, if someone calls it twice it only expects created list

vkrasnovyd: Could you please explain the potential problem in more detail?

ArsenPidhoretskyi: image
What do you expect from this code?
I expect that both calls return me a list with on argument. But your code will return the first time one and second time two

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #510, opened 2023-04-27
'Solution'

VTeteruk:


app/main.py
```diff
@@ -1,8 +1,31 @@
...
+
+    for person in all_people:
+        for status in ("wife", "husband"):
+            if status in person.__dict__:
+                if person.__dict__[status] is None:
+                    del person.__dict__[status]
+                else:
+                    person.__dict__[status] = person.people[
+                        person.__dict__[status]
+                    ]
```
ArsenPidhoretskyi, line 29: So hard code....
Please simplify it.
You can do just in two simple loops:
1. Create all people
2. Set wife/human due to person( if person.get("wife") # than set Person.people[person["name"]] = wife(get from Person)

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,18 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list:
```
vasylhnatiuk, line 10:
```suggestion
def create_person_list(people: list[dict]) -> list[Person]:
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #524, opened 2023-05-11
'Solution'

Rovalexx:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        elif ("husband" in person) and (person["husband"] is not None):
+            new_person.husband = person["husband"]
+        else:
+            pass
+        person_list.append(new_person)
+    for person in person_list:
+        if hasattr(person, "wife"):
+            person.wife = Person.people[person.wife]
+        elif hasattr(person, "husband"):
+            person.husband = Person.people[person.husband]
```
olena-hul, line 27: Use first loop to create person instances and second loop - to fill wie/husband attributes

olena-hul (changes requested):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+        person_list.append(new_person)
+
+    for person in person_list:
```
ArsenPidhoretskyi, line 18: Remove this loop. No need to use loop inside loop, you need only inner loop.
How objects works? By value or by link?

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        for human in people:
+            if person.name == human["name"]:
+                if ("wife" in human) and (human["wife"] is not None):
```
ArsenPidhoretskyi, line 21: Use .get: human.get("wife") is not None
The same with husband

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+        person_list.append(new_person)
```
olena-hul, line 16: Use list comprehension

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+
+    for human in people:
+        if human.get("wife") is not None and human["wife"] is not None:
```
olena-hul, line 19:
```suggestion
        if human.get("wife") is not None:
```

The second part of check is almost identical to the first

olena-hul: The same with husband

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
+        elif human.get("husband") is not None and human["husband"] is not None:
+            Person.people[human["name"]].husband =\
+                Person.people[human["husband"]]
```
olena-hul, line 23: Use variables, don't make such long lines

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #525, opened 2023-05-12
solution

uzlss:
I'm not sure, is it the best solution, I would like to know what actually is...

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -8,7 +8,7 @@ class Person:
     ) -> None:
         self.name = name
         self.age = age
-        self.__class__.people.update({self.name: self})
+        self.people[self.name] = self
 
 
 def create_person_list(people: list[dict]) -> list[Person]:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    people = {}
 
+    def __init__(
+            self,
+            name: str,
+            age: int,
+    ) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people.update({self.name: self})
```
ArsenPidhoretskyi, line 11: No need to use `__class__` also you can set by [] instead of method update

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+def create_person_list(people: list[dict]) -> list[Person]:
+    person_list = [
+        Person(
+            person.get("name"),
+            person.get("age")
+        )
+        for person in people
+    ]
+    for num, person in enumerate(people):
+        for key, value in person.items():
```
ArsenPidhoretskyi, line 23: This loop is not needed
Just two ifs:
1. if person.get("wife") is not None
2. if person.get("husband") is not None

ArsenPidhoretskyi (changes requested):

app/main.py
uzlss: should I add "if" into list comprehension? I think, it will cause error because of 4 arguments given instead of 3, or there is another possibility for it?

vasylhnatiuk (approved):

PR #527, opened 2023-05-15
Solution

FadieievaA:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
-    # write your code here
-    pass
+    for human in people:
+        Person.people[human["name"]] = Person(human["name"], human["age"])
+        if human.get("husband"):
+            wife = Person.people[human["name"]]
+            husband = Person.people[human["husband"]]
+            wife.husband = husband
+            husband.wife = wife
+    return [person for name, person in Person.people.items()]
```
ArsenPidhoretskyi, line 18: This works wrong
Because:
1. You must return only created list, not all of items
2. You must set wife also(for this you must check if human.get("wife")
3. Person.people[human["name"]] = Person(human["name"], human["age"])
Is so strange because it do the same inside init method
4. Use 2 loops: 1 - create all person, 2 - set wife/husband

ArsenPidhoretskyi (changes requested):

app/main.py
FadieievaA: Thank you.

For the p.2 I assumed that if someone in the list has husband, I cross-link wife-husband and therefore I don't need to check "wife" key

PR #530, opened 2023-05-17
Create_person_list implemented

thorium-zero:


app/main.py
```diff
@@ -1,8 +1,31 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    spouse_types: set[str] = {"wife", "husband"}
+    for person in people:
+        if person["name"] in Person.people:
+            continue
+        current_person = Person(person["name"], person["age"])
+        Person.people[person["name"]] = current_person
```
YuriiKindrat, line 16: why do u need this ? the person will be added to dict in init method

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+        spouse_type: set = (spouse_types & person.keys())
+        if len(spouse_type) < 1 or person.get(next(iter(spouse_type))) is None:
```
YuriiKindrat, line 18: u can use pop method instead of iter + next to write less code ) the same is for other cases

app/main.py
```diff
@@ -1,8 +1,31 @@
 class Person:
-    # write your code here
-    pass
+    people: dict = {}
```
YuriiKindrat, line 2: no need to write type annotation for such cases))

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+            continue
+        current_person = Person(person["name"], person["age"])
+        Person.people[person["name"]] = current_person
+        spouse_type: set = (spouse_types & person.keys())
+        if len(spouse_type) < 1 or person.get(next(iter(spouse_type))) is None:
+            continue
+        spouse_name: str = person.get(next(iter(spouse_type)))
+        if spouse_name not in Person.people:
+            spouse_info = next((p for p in people
+                                if p["name"] == spouse_name), None)
```
YuriiKindrat, line 23: "p" is a bad name for variable

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+            Person.people[spouse_name] = Person(spouse_info["name"],
+                                                spouse_info["age"])
```
YuriiKindrat, line 25: the same as before, why do u need this ?

YuriiKindrat (changes requested):

app/main.py
thorium-zero: Agree, that action is redundant

app/main.py
thorium-zero: I disagree, that will remove an item from my template set. I don't want to do that

app/main.py
thorium-zero: Agreed, this operation is redundant

app/main.py
LiudmylaKulinchenko: It's still a best practice to use more descriptive variable name here

LiudmylaKulinchenko: But it doesn't have to be `current_person`, it is really too long. You can use just `person`

app/main.py
```diff
@@ -1,8 +1,31 @@
...
-    pass
+
+def create_person_list(people: list) -> list[Person]:
+    spouse_types: set[str] = {"wife", "husband"}
+    for person in people:
+        if person["name"] in Person.people:
+            continue
+        current_person: Person = Person(person["name"], person["age"])
+        spouse_type: set = (spouse_types & person.keys())
+        if len(spouse_type) < 1 or person.get(next(iter(spouse_type))) is None:
```
LiudmylaKulinchenko, line 17: 1. You can check `if not spouse_type` instead of len < 1
2. You don't need this check because from the solutioun you know there only will be valid dicts
3. Try to remove this next(iter()). You should care to make your code readable, it is hard to read now

app/main.py
YuriiKindrat: u can make in 16th line like this 
`spouse_type: str = (spouse_types & person.keys()).pop()`
then u can use this variable in ur code 
in 29th line u can use `remove` method instead of `-`
this will make u code DRY - cause now u are making the same thing every time u are needed of spouse_type
why don't u create a variable for this ?

YuriiKindrat (changes requested):

LiudmylaKulinchenko (changes requested):

app/main.py
thorium-zero: that's a valid point, but semantically seems a bit weird. Looks like python is still chinese to me :(

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    for person_info in people:
+        Person(person_info["name"], person_info["age"])
+    for person_info in people:
+        person = Person.people[person_info["name"]]
+        if "wife" in person_info and person_info["wife"] is not None:
```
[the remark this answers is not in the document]
thorium-zero: yeah, this implementation is MUCH worse performance-wise than the previous one though (if you'd work with some large database of people that is, but to hell with that already...).

YuriiKindrat (approved):

app/main.py
```diff
@@ -1,8 +1,31 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list) -> list[Person]:
+    spouse_types: set[str] = {"wife", "husband"}
+    for person in people:
+        if person["name"] in Person.people:
+            continue
+        current_person: Person = Person(person["name"], person["age"])
```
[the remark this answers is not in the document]
thorium-zero: It does. When you have your variable type annotated you also have the intelli-sense for it

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list) -> list[Person]:
+    spouse_types: set[str] = {"wife", "husband"}
```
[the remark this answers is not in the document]
thorium-zero: yeah, it is unnecessary alright, but I don't want my code be full of "Any"-type hints for my variables. It is bad enough that python has no static type checking and metadata availability, but at least I want to have some intelli-sense on my vars

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+    for person in people:
+        if person["name"] in Person.people:
+            continue
+        current_person: Person = Person(person["name"], person["age"])
+        spouse_type: set = (spouse_types & person.keys())
```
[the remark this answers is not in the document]
thorium-zero: i second that. it's probably more readable approach to use that set method instead

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+        if len(spouse_type) < 1 or person.get(next(iter(spouse_type))) is None:
+            continue
+        spouse_name: str = person.get(next(iter(spouse_type)))
```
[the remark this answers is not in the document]
thorium-zero: yeah, good point, I forgot about that method

PR #536, opened 2023-05-20
implemented class person and function create_persons_list

Oknametra:
I'm not sure about the return type "Person" and @classmethod, maybe I don't really need them and we can optimize it

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,36 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
+
+    @classmethod
+    def get_person(cls, name: str) -> "Person":
+        return cls.people.get(name)
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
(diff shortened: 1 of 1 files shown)


Oknametra:
maybe I really made it too complicated

app/main.py
```diff
@@ -1,8 +1,36 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
Vasyl-Smutok, line 7:
```suggestion
        self.people[name] = self
```

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+
+    @classmethod
+    def get_person(cls, name: str) -> "Person":
```
Vasyl-Smutok, line 10:
```suggestion
    def get_person(cls, name: str) -> Person:
```

Vasyl-Smutok (changes requested):

app/main.py
Oknametra: >
after this change ERROR tests/test_main.py - NameError: name 'Person' is not defined

Vasyl-Smutok (approved):
Nice 👍

PR #538, opened 2023-05-24
done

chipapapo:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+    for index, person in enumerate(people):
+        try:
+            wife = Person.people[person["wife"]]
+            list_of_completed_people[index].wife = wife
+        except KeyError:
+            try:
+                husband = Person.people[person["husband"]]
+                list_of_completed_people[index].husband = husband
+            except KeyError:
+                pass
```
ArsenPidhoretskyi, line 22: For now it works so hard....
You can use if condition:
`if person.get("wife") is not None:`

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_completed_people = [Person(i["name"], i["age"]) for i in people]
```
vasylhnatiuk, line 11: `i` is a bad name

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #544, opened 2023-05-29
done

maccarets:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        temp = Person(person["name"], person["age"])
+        result.append(temp)
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+
+        if person.get("husband"):
+            Person.people[person["name"]].husband = \
+                Person.people[person["husband"]]
```
YuriiKindrat, line 22: no need to use linebreaker here, u can break the line through the brackets

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for person in people:
+        temp = Person(person["name"], person["age"])
+        result.append(temp)
```
YuriiKindrat, line 14: gj, but why do u need this variable here? can u add person to the list without creating variable ?

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Good job 🐥

PR #545, opened 2023-05-29
Solved

oleksiikolii:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        return cls.people.get(name)
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        name = person["name"]
+        age = person["age"]
```
YuriiKindrat, line 18: no need to create variables here, u can get name directly while creating Person

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        Person(name, age)
+    for person in people:
+        new_person = Person.get_person(person["name"])
+        if "wife" in person and person["wife"] is not None:
+            new_person.wife = Person.people[person["wife"]]
+            Person.people[person["wife"]].husband = new_person
+        if "husband" in person and person["husband"] is not None:
+            new_person.husband = Person.people[person["husband"]]
```
YuriiKindrat, line 26: Use just dict.get() method in the condition, instead of this long one

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        Person(person["name"], person["age"])
+    for person in people:
+        new_person = Person.get_person(person["name"])
+        if person.get("wife") is not None:
```
YuriiKindrat, line 20: no need to write `is not None` here, condition won't be True if get method will return None

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+            new_person.wife = Person.people[person["wife"]]
+            Person.people[person["wife"]].husband = new_person
+        if person.get("husband") is not None:
+            new_person.husband = Person.people[person["husband"]]
```
YuriiKindrat, line 24: will ur code work without this part ? u are adding husband for wife and wife for husband in code before

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Good job 🎉

PR #546, opened 2023-05-29
Solution

DmytroTsehelnyk:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    person_list = []
+
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+        person_list.append(new_person)
+
+    for person in people:
+        new_person = Person.people[person["name"]]
+
+        if "wife" in person and person["wife"]:
```
YuriiKindrat, line 22: Use just dict.get() method in the condition, instead of this long on

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #548, opened 2023-05-29
'Solution'

ant-komarov:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        person_list.append(Person(person["name"], person["age"]))
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```
YuriiKindrat, line 16: Use just dict.get() method in the condition, instead of this long one

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        if "husband" in person and person["husband"] is not None:
+            Person.people[person["name"]].husband =\
```
YuriiKindrat, line 19: try to avoid using linebreakers, u can split this line through brackets

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #549, opened 2023-05-29
Solution

AntonShpakovych:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    _ = [Person(human["name"], human["age"]) for human in people]
```
YuriiKindrat, line 11: it is better to use for loops for creating smth, list comprehension is used for creating list

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+
+    for human in people:
+        if human.get("wife"):
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
+        elif human.get("husband"):
+            Person.people[human["name"]
+                          ].husband = Person.people[human["husband"]]
```
YuriiKindrat, line 18: bad formating :(

YuriiKindrat (changes requested):

app/main.py
AntonShpakovych: Yep, my mistake)

app/main.py
AntonShpakovych: It was auto-formatting in my editor. I think I need to consider disabling this feature or switch to PyCharm)

AntonShpakovych: Can you suggest a better  auto-formatting extension for Visual Studio Code. If u know?

app/main.py
```diff
@@ -8,12 +8,13 @@ def __init__(self, name: str, age: int) -> None:
...
+        Person(human["name"], human["age"])
 
     for human in people:
         if human.get("wife"):
             Person.people[human["name"]].wife = Person.people[human["wife"]]
         elif human.get("husband"):
-            Person.people[human["name"]
-                          ].husband = Person.people[human["husband"]]
+            Person.people[human["name"]].husband = \
+                Person.people[human["husband"]]
```
YuriiKindrat, line 19: it is better to avoid using linebreakers 
u can break this line through the brackets, but with better formating )

YuriiKindrat (changes requested):

app/main.py
AntonShpakovych: In PyCharm this problem fixed automatically or no? Like if I use PyCharm instead of VSCode i won't be get this bad formatting?

AntonShpakovych: Because I have already installed Black but for him this rows not too long, and I think I must change  maybe max_length_of_row in Black to maybe flake8 max_length_of_row. But i think its not good idea because max_length_of_row depends to project style where u work, not sure)

app/main.py
```diff
@@ -1,8 +1,20 @@
...
-    # write your code here
-    pass
+    for human in people:
+        Person(human["name"], human["age"])
+
+    for human in people:
+        if human.get("wife"):
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
+        elif human.get("husband"):
+            Person.people[human["name"]].husband = Person.people[
```
LiudmylaKulinchenko, line 18: Create additional variable to avoid bad formatting

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for human in people:
+        Person(human["name"], human["age"])
+
+    for human in people:
+        if human.get("wife"):
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
```
LiudmylaKulinchenko, line 16: Create here `wife` variable as well to have the same code style

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    for human in people:
+        Person(human["name"], human["age"])
+
+    for human in people:
+        if human.get("wife"):
+            wife = Person.people[human["wife"]]
+            Person.people[human["name"]].wife = wife
+        elif human.get("husband"):
+            husband = Person.people[human["husband"]]
+            Person.people[human["name"]].husband = husband
```
Oleg-s-s, line 20:
```suggestion
for human in people:
    person = Person.people[human["name"]]
    if human.get("wife"):
        person.wife = Person.people[human["wife"]]
    elif human.get("husband"):
        person.husband = Person.people[human["husband"]]
```

Oleg-s-s (approved):
What do you think about the comment? :upside_down_face:

PR #550, opened 2023-05-29
Solution

volodymyr-komarnyckyi:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
+        if "wife" in item:
+            wife_name = item["wife"]
+            if wife_name in Person.people:
+                person.wife = Person.people[wife_name]
+                person.wife.husband = person
+        if "husband" in item:
+            husband_name = item["husband"]
+            if husband_name in Person.people:
+                person.husband = Person.people[husband_name]
+                person.husband.wife = person
```
YuriiKindrat, line 24: why do u need nested condintions here? u can use `and` for this, also u can use `get ` metod to get wife/husband from dict)

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for item in people:
```
LiudmylaKulinchenko, line 13: use more descriptive variable name, not `item`

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for people_info in people:
```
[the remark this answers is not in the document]
volodymyr-komarnyckyi: Thanks for feedback

LiudmylaKulinchenko (approved):
Good job 🚀😎

PR #551, opened 2023-05-29
Solution

diana-shyrokikh:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
-    pass
+    persons = []
+
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+
+        if "wife" in person and person["wife"] is not None:
+            new_person.wife = person["wife"]
+        elif "husband" in person and person["husband"] is not None:
+            new_person.husband = person["husband"]
```
YuriiKindrat, line 20: gj, but use just dict.get() method in the condition, instead of this long one

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #553, opened 2023-05-29
Solution

YKonoplyov:


app/main.py
```diff
@@ -1,8 +1,32 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person in people:
+        (Person(name=person["name"], age=person["age"]))
+    print(Person.people)
```
YuriiKindrat, line 15: remove unnecessary prints from ur code

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person in people:
+        (Person(name=person["name"], age=person["age"]))
```
YuriiKindrat, line 14: unnecessary brackets here

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+            Person.people[person["name"]].husband =\
+                Person.people[person["husband"]]
+            person_list.append(Person.people[person["name"]])
+
+        elif "wife" in person and person["wife"] is not None:
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+            person_list.append(Person.people[person["name"]])
+
+        else:
+            person_list.append(Person.people[person["name"]])
```
YuriiKindrat, line 29: why don't u add persons to the list using first for loop ?

app/main.py
```diff
@@ -1,8 +1,32 @@
...
-    pass
+    person_list = []
+
+    for person in people:
+        (Person(name=person["name"], age=person["age"]))
+    print(Person.people)
+
+    for person in people:
+
+        if "husband" in person and person["husband"] is not None:
```
YuriiKindrat, line 19: Use just dict.get() method in the condition, instead of this long one

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+            Person.people[person["name"]].husband =\
+                Person.people[person["husband"]]
```
YuriiKindrat, line 21: try to avoid using linebreakers, u can split a string through brackets

YuriiKindrat (changes requested):

app/main.py
YKonoplyov: May be I`m not right, but "for" loop is above?

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+    person_list = []
+
+    for person in people:
+        Person(name=person["name"], age=person["age"])
+
+    for person in people:
+
+        if "husband" in person and person.get("husband") is not None:
+            Person.people[
+                person["name"]].husband = Person.people[person["husband"]]
```
LiudmylaKulinchenko, line 20: Braces should be located in one of two options: open and start with the text or have a line break between the text. You have both options now

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+            person_list.append(Person.people[person["name"]])
+
+        elif "wife" in person and person.get("wife") is not None:
+            Person.people[
+                person["name"]].wife = Person.people[person["wife"]]
```
LiudmylaKulinchenko, line 25: Split on variables here as well

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+        Person(name=person["name"], age=person["age"])
+
+    for person in people:
+
+        if "husband" in person and person.get("husband") is not None:
+            Person.people[
+                person["name"]].husband = Person.people[person["husband"]]
+            person_list.append(Person.people[person["name"]])
+
+        elif "wife" in person and person.get("wife") is not None:
```
LiudmylaKulinchenko, line 23: `if` statement will be True here if key exists, `is not None` redundant here

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+            Person.people[
+                person["name"]].wife = Person.people[person["wife"]]
+            person_list.append(Person.people[person["name"]])
+
+        else:
```
LiudmylaKulinchenko, line 28: Are you sure you need `else` here, according to the task description?

app/main.py
LiudmylaKulinchenko: Try to split on variables, it would be even better

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 😎

PR #555, opened 2023-05-30
Solution

YevheniiIovchu:


app/main.py
```diff
@@ -1,8 +1,29 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```

app/main.py
```diff
@@ -1,8 +1,29 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person = Person(name, age)
+        person_list.append(person)
```
YuriiKindrat, line 16: try to do this in one line

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+
+    for person_data in people:
+        person = Person.people[person_data["name"]]
+        if person_data.get("wife") is not None:
```
YuriiKindrat, line 20: no need to write `is not None` here

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+            wife_name = person_data["wife"]
+            person.wife = Person.people[wife_name]
+            Person.people[wife_name].husband = person
+        if person_data.get("husband") is not None:
+            husband_name = person_data["husband"]
+            person.husband = Person.people[husband_name]
+            Person.people[husband_name].wife = person
```
YuriiKindrat, line 27: will code work without this part ? u are creating wife for husband and husband for wife before )

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person = Person(name, age)
+        person_list.append(person)
+
+    for person_data in people:
+        person = Person.people[person_data["name"]]
+        if person_data.get("wife") is not None:
+            wife_name = person_data["wife"]
```
YuriiKindrat, line 21: could u create this variable before if condition ? )

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_data in people:
```
LiudmylaKulinchenko, line 12: Previous comment was about you can use list comprehension here 🤫

LiudmylaKulinchenko (changes requested):

PR #557, opened 2023-05-30
Solution

erikagayan:


app/main.py
```diff
@@ -1,8 +1,32 @@
...
+            if wife_name in Person.people:
+                person.wife = Person.people[wife_name]
+                person.wife.husband = person
+
+        if "husband" in pers:
+            husband_name = pers["husband"]
+
+            if husband_name in Person.people:
+                person.husband = Person.people[husband_name]
+                person.husband.wife = person
```
YuriiKindrat, line 28: could u create wife_name\husband_name vars before ur if condition using get method, so u could use only one if condition instead of nested conditions?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for pers in people:
```
YuriiKindrat, line 13: pers is a bad name for variable) use more descriptive naming

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done 🎈

PR #559, opened 2023-05-30
'Solution'

yakhlamova:


app/main.py
```diff
@@ -1,8 +1,32 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for individ in people:
+        name = individ["name"]
+        age = individ["age"]
+        person = Person(name, age)
+        person_list.append(person)
```
YuriiKindrat, line 17: could u make this in one line ?

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+
+    for individ in people:
+        name = individ["name"]
```
YuriiKindrat, line 20: u can make this inside `Person.people[]` no need to create var for this

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+        person = Person.people[name]
+        if "wife" in individ and individ["wife"] is not None:
```
YuriiKindrat, line 22: Use just dict.get() method in the condition, instead of this long one

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+            wife_name = individ["wife"]
+            person.wife = Person.people[wife_name]
+            Person.people[wife_name].husband = person
+
+        if "husband" in individ and individ["husband"] is not None:
+            husband_name = individ["husband"]
+            person.husband = Person.people[husband_name]
+            Person.people[husband_name].wife = person
```
YuriiKindrat, line 30: will ur code work without this part ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for individ in people:
```
LiudmylaKulinchenko, line 14: Try to rewrite in one line. Hint: list comprehention 🤫

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        person_list.append(Person(individ["name"], individ["age"]))
+
+        if individ.get("wife") in Person.people:
+            Person.people[individ["name"]].wife \
```
LiudmylaKulinchenko, line 18: Don't use `= \` , create temporary variable instead or use parentheses

LiudmylaKulinchenko (changes requested):

app/main.py
yakhlamova: Could you please a bit clarify because I don`t understand. Do I need to rewrite the whole for loop in one line?

YuriiKindrat: Why do u need to create a blank list `person_list` and then make a separate for loop for creating an instances of People class
U can use list comprehension here for list + instances creating )

app/main.py
```diff
@@ -1,8 +1,22 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for individ in people:
+        person_list.append(Person(individ["name"], individ["age"]))
+
+        if individ.get("wife") in Person.people:
+            wife = individ["wife"]
```
YuriiKindrat, line 18: don't repeat ur self, u can create a variable to avoid gettign value from dict twice here)

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    person_list = [Person(indiv["name"], indiv["age"]) for indiv in people]
```
YuriiKindrat, line 13: indiv is a bad name for wariable, use full names for naming

YuriiKindrat (approved):

app/main.py
yakhlamova: I had a flake8 issue with the line being too long and it was like in one or two symbols, so I decided just too change the name. Thank u for comment I won`t do this again.

PR #560, opened 2023-05-30
'Solution'

RedCatHot:


app/main.py
```diff
@@ -1,8 +1,29 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = []
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+
+        married_key = "wife" if "wife" in person else "husband"
+        married_name = person.get(married_key)
+        married = Person.people.get(married_name)
```
LiudmylaKulinchenko, line 17: Try to rename this variables, the names do not correspond to what is in them

RedCatHot: How I should rename variables? We are talking about married_key...?
Is "spouse" instead "married" okay?

LiudmylaKulinchenko: Yes, it is great!

LiudmylaKulinchenko (approved):
Well done 🌼

PR #561, opened 2023-05-30
Solution

imelnyk007:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        person_list.append(Person(name, age))
+
+    for person in people:
+        if "wife" in person:
+            wife = person["wife"]
+            if wife in Person.people:
+                Person.people[person["name"]].wife = Person.people[wife]
+        if "husband" in person:
+            husband = person["husband"]
+            if husband in Person.people:
```
LiudmylaKulinchenko, line 24: it's better to use `and` instead of nested `if` blocks

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        name = person["name"]
+        age = person["age"]
+        person_list.append(Person(name, age))
+
+    for person in people:
+        if "wife" in person and person["wife"] in Person.people:
```
Oleg-s-s, line 18: You can just use dict.get() method in the condition

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        if "husband" in person and person["husband"] in Person.people:
+            Person.people[person["name"]].husband = Person.people[
+                person["husband"]
+            ]
```
Oleg-s-s, line 23: Think about creating additional variable for Person.people[person["name"]], this will make your code look cleaner

Oleg-s-s (changes requested):

Oleg-s-s (approved):
Looks great)

PR #562, opened 2023-05-30
Solution

Kirontiko:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
```
LiudmylaKulinchenko, line 12: Use list comprehension in such cases

LiudmylaKulinchenko (changes requested):

PR #563, opened 2023-05-30
implemented py-person-class

mariiahorbova:


app/main.py
```diff
@@ -1,8 +1,29 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int, **kwargs) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        age = person["age"]
+        new_person = Person(name, age)
+        person_list.append(new_person)
+
+    for person in people:
+        name = person["name"]
+        wife = person.get("wife")
+        husband = person.get("husband")
+
+        if wife is not None:
```
LiudmylaKulinchenko, line 23: `is not None` redundant here, since in your case if statement will be True if key exists

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+            Person.people[name].wife = Person.people[wife]
+
+        if husband is not None:
```
LiudmylaKulinchenko, line 26: Same here about `is not None`

LiudmylaKulinchenko (changes requested):

app/main.py
mariiahorbova: done

app/main.py
mariiahorbova: done

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
```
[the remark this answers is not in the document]
mariiahorbova: done

PR #569, opened 2023-05-30
Solution for Person Class

OlhaTryhub:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list_obj = [Person(pers["name"], pers["age"]) for pers in people]
+    for i in range(len(people)):
```
LiudmylaKulinchenko, line 12: Try to iterate by element in people, not by index

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        if people[i].get("wife"):
+            people_list_obj[i].wife = Person.people[people[i]["wife"]]
+        if people[i].get("husband"):
+            people_list_obj[i].husband\
```
LiudmylaKulinchenko, line 16: Don't use ` \` as a linebreaker, create temporary variable instead or use parentheses

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list_obj = [Person(pers["name"], pers["age"]) for pers in people]
+    for pers in people:
```
YuriiKindrat, line 12: gj, but it is better to use full names for vars naming, person - not pers

YuriiKindrat (approved):

PR #570, opened 2023-05-30
Solution

Po-SvinSkipodla:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        persons_ls.append(Person(person["name"], person["age"]))
+
+    for person in people:
+        current_person = Person.people[person["name"]]
+
+        if "wife" in person.keys() and person["wife"]:
+            current_person.wife = Person.people[person["wife"]]
+
+        if "husband" in person.keys() and person["husband"]:
+            current_person.husband = Person.people[person["husband"]]
```
Oleg-s-s, line 23: You can use dict.get() method here to simplify checking

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons_ls = []
+
+    for person in people:
```

LiudmylaKulinchenko (changes requested):

PR #575, opened 2023-05-31
Solution

anastmishchuk:


app/main.py
```diff
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
```
vsmutok, line 7:
```suggestion
        self[self.name] = self
```

app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        name = person.get("name")
+        age = person.get("age")
+        human = Person(name, age)
+        person_list.append(human)
```
vsmutok, line 16: try to use list comprehensionhere

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    for partner in people:
+        if partner.get("wife") is not None:
+            name_partner = partner.get("wife")
+            Person.people[partner["name"]].wife\
```
vsmutok, line 20: better to use `()` then `\`

vsmutok (changes requested):

anastmishchuk (commented):
When I changed the 7th string to `self[self.name] = self`, it was an error so I changed it to `self.people[self.name] = self.` Other requested changes I made as proposed.

vasylhnatiuk (approved):

PR #579, opened 2023-06-03
Solution

Anastasia-Su:


vasylhnatiuk (commented):
You wrote complicated code and you used so many loops. The simple way to resolve this will be to use the first loop only for creating a person instance and the second loop with also pelople (check and set).

vasylhnatiuk (changes requested):
rewrite your code pls

Anastasia-Su:
Rewrote it. Seems now it looks simplier. Unfortunately, I don't know how to simplify it further. Could you please advise?

app/main.py
```diff
@@ -1,8 +1,45 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for human in people:
+        new_person = Person(name=human["name"], age=human["age"])
+
+        Person.people.update({human["name"]: new_person})
```
ArsenPidhoretskyi, line 15: No, you need to set this inside init method

app/main.py
```diff
@@ -1,8 +1,45 @@
...
+                if instance.name == husband["husband"]
+            ]
+
+            filt_husband_person[0].wife = filt_wife_person[0] \
+                if filt_wife_person[0] \
+                else wife["wife"]
+
+            filt_wife_person[0].husband = filt_husband_person[0] \
+                if filt_husband_person[0] \
+                else husband["husband"]
```
ArsenPidhoretskyi, line 43: Replace this by one loop:
Iterate for each person and check for wife/husband inside person

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
The same with husband

PR #581, opened 2023-06-04
Solution0506

MichaelKyiv:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for pers in people:
```
ArsenPidhoretskyi, line 12: pers is not good name
rename to person

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: @MichaelKyiv not fixed

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #588, opened 2023-06-13
Solution

NazarLytvyniuk:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
-    pass
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
+
+
+def create_person_list(people: list[dict]) -> list:
+    for person in people:
+        _ = Person(name=person.get("name"), age=person.get("age"))
```
Dimosphen1, line 13: There is no need to process the None output of creating a person, you can omit that and leave just initialization.

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        person_name = person.get("name")
+
+        if (wife := person.get("wife")) is not None:
+            Person.people[person_name].wife = Person.people[wife]
+            continue
+
+        if (husband := person.get("husband")) is not None:
+            Person.people[person_name].husband = Person.people[husband]
+
+    return [person for person in Person.people.values()]
```
Dimosphen1, line 25: Try not to use a list comprehension here, there is a way to avoid iteration

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+
+
+def create_person_list(people: list[dict]) -> list:
+    for person in people:
+        _ = Person(name=person.get("name"), age=person.get("age"))
+
+    for person in people:
+        person_name = person.get("name")
+
+        if (wife := person.get("wife")) is not None:
```
Dimosphen1, line 18: There is no need for `is not None`, the assignment expression would provide that check

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+            Person.people[person_name].wife = Person.people[wife]
+            continue
+
+        if (husband := person.get("husband")) is not None:
```
Dimosphen1, line 22: There is no need for `is not None`, the assignment expression would provide that check

app/main.py
Dimosphen1, line 25: Do not forget to leave one blank line at the end of the file

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
+
+
+def create_person_list(people: list[dict]) -> list:
+    for person in people:
+        _ = Person(name=person.get("name"), age=person.get("age"))
+
+    for person in people:
```
Dimosphen1, line 15: Try to avoid the second iteration through the `people` list and collect information regarding wife and husband names throughout the first iteration. You can be more targeted in the second iteration when assigning attributes to the class.

Dimosphen1 (changes requested):

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #598, opened 2023-06-26
Solution

ivanzh13:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        if person.get("wife"):
+            temp_person.wife = Person.people.get(person["wife"])
+            if temp_person.wife:
+                temp_person.wife.husband = temp_person
+        if person.get("husband"):
+            temp_person.husband = Person.people.get(person["husband"])
+            if temp_person.husband:
+                temp_person.husband.wife = temp_person
+        person_list.append(temp_person)
+    return person_list
```
ArsenPidhoretskyi, line 23: Do this in tow loop and simplify code.
In first loop create all Person
In second loop set wife/husband

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.age = age
+        self.__class__.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = list()
+    for person in people:
+        _ = Person(person["name"], person["age"])
```
vasylhnatiuk, line 13:
```suggestion
       person_list.append(Person(person["name"], person["age"]))
```

app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+    people = dict()
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
vasylhnatiuk, line 7:
```suggestion
        self.people[name] = self
```

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    person_list = list()
+    for person in people:
+        _ = Person(person["name"], person["age"])
+    
+    for person in people:
+        if person.get("wife"):
+            Person.people.get(person["name"]).wife = Person.people.get(person["wife"])
+        if person.get("husband"):
+            Person.people.get(person["name"]).husband = Person.people.get(person["husband"])
+    return list(Person.people.values())
```
vasylhnatiuk, line 20:
```suggestion
    return person_list
```

app/main.py
vasylhnatiuk: it will be better way to return list

vasylhnatiuk (changes requested):
image

app/main.py
ivanzh13: Explain why if we need add to class's attribute we use instance of class not class

vasylhnatiuk: sorry, my mistake, but it would be better to use Person.people

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #601, opened 2023-06-26
all tests passed

bkirichenko:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        name = person.get("name")
+        age = person.get("age")
+        if name is not None and age is not None:
+            person_list.append(Person(name, age))
```
vsmutok, line 17: use list comprehension

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
+            wife = Person.people[person["wife"]]
```
vsmutok, line 20: better to use `get`
fix everywhere 
```suggestion
        if "wife" in person and person.get("wife") is not None:
            wife = Person.people[person["wife"]]
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,8 +1,24 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [
+        Person(person.get("name"), person.get("age"))
+        for person in people
+        if person.get("name") is not None and person.get("age") is not None
+    ]
+    for person in people:
+        if "wife" in person and person.get("wife") is not None:
```
ArsenPidhoretskyi, line 18: Just do like this:
```suggestion
        if person.get("wife") is not None:
```
With husband the same

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done! But usage .get inside if becomes redundant. We already know that this key present

PR #611, opened 2023-06-30
Solution

hrabovii:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
-    # write your code here
-    pass
+    person_instances = []
+    for person_dict in people:
+        person_instance = Person(person_dict["name"], person_dict["age"])
+        person_instances.append(person_instance)
+
+    for person_dict in people:
+        person = Person.people[person_dict["name"]]
+        if "wife" in person_dict and person_dict["wife"] is not None:
```
ArsenPidhoretskyi, line 18: You can use .get instead of two conditions:
person_dict.get("wife") is not None
The same with husband

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_instances = []
+    for person_dict in people:
+        person_instance = Person(person_dict["name"], person_dict["age"])
+        person_instances.append(person_instance)
```
vsmutok, line 14: use list comprehension here

vsmutok (changes requested):

vsmutok (approved):
Good Job 👍

PR #623, opened 2023-07-10
person class

WoyDiam:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
-    pass
+    person_list = []
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person = Person(name, age)
+        person_list.append(person)
+    for person_data in people:
+        person = Person.people[person_data["name"]]
+        if "wife" in person_data and person_data["wife"] is not None:
```
ArsenPidhoretskyi, line 22: Simplify condition and just use .get:
person_data.get("wife") is not None
Fix the same with husband

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,34 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+
+        self.__class__.people[name] = self
```
Grayder0152, line 8:
```suggestion
        self.people[name] = self
```

app/main.py
```diff
@@ -1,8 +1,34 @@
...
-    pass
+    person_list = []
+    person_dict = {}
+
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person = Person(name, age)
+        person_list.append(person)
+        person_dict[name] = person
```
Grayder0152, line 23: why did you create this dict ?

Grayder0152 (changes requested):

app/main.py
WoyDiam: To use the get() method. I couldn't pass one test without it.

app/main.py
```diff
@@ -1,8 +1,18 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
```
ArsenPidhoretskyi, line 7: FYI, better to use:
self.people[name] = self

ArsenPidhoretskyi (approved):
Is looks so cool)

PR #626, opened 2023-07-13
Solution

AndriiIshchenko:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(human["name"], human["age"]) for human in people]
+    for human in people:
+        if human.get("wife"):
+            husband = Person.people[human["name"]]
+            wife = Person.people[human["wife"]]
+            husband.wife = wife
+            wife.husband = husband
```
ArsenPidhoretskyi, line 17: nothing said that if "wife" we need to set husband. Add additional check for husband. We need to set values due to input data

ArsenPidhoretskyi (changes requested):

app/main.py
vasylhnatiuk: It is logical how you wrote it, but more readable code will be how Arsen suggested it. Rewrite please

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,8 +1,19 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(human["name"], human["age"]) for human in people]
+    for human in people:
+        if human.get("wife"):
+            Person.people[human["name"]].wife = Person.people[human["wife"]]
+        if human.get("husband"):
+            Person.people[human["name"]].husband \
+                = Person.people[human["husband"]]
```
Grayder0152, line 17: use `()` instead of `\`

Grayder0152 (changes requested):

app/main.py
AndriiIshchenko: str2
Pycharm raises error

AndriiIshchenko: if you can just show how to do it right way

vsmutok (approved):
Good Job 👍

PR #635, opened 2023-07-23
Solution

AlexPelevin99:


pavlejviki (changes requested):
All checks should have been passed.

AlexPelevin99:
one checks not passed because test has error
image

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        self.age = age
+        self.people[self.name] = self
+
+    @classmethod
+    def find_by_name(cls, name: str) -> Person | None:
+        return cls.people[name]
+
+
+def create_person_list(people: list) -> list[Person]:
+    for p in people:
```
Dimosphen1, line 18: Such variable name as `p` is not allowed as stated by `flake8`. Replace it with a more descriptive one, e.g. `person`.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        Person(p["name"], p["age"])
 
+    for p in people:
```
Dimosphen1, line 21: Do not use single-letter variable names here.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    people: dict = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
+
+    @classmethod
+    def find_by_name(cls, name: str) -> Person | None:
+        return cls.people[name]
```
Dimosphen1, line 14: It would be better to use `.get()` here thus making code less error-prone.

Dimosphen1 (changes requested):
Resolve issues raised by `flake8`.

app/main.py
```diff
@@ -1,8 +1,29 @@
+from __future__ import annotations
```
Dimosphen1, line 1: Do not use this import here, it breaks the tests.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    people: dict = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
+
+    @classmethod
+    def find_by_name(cls, name: str) -> Person | None:
+        return cls.people.get(name)
```
Dimosphen1, line 14: There is no need for this classmethod, you should get value from `people` attribute from the `Person` class directly.

Dimosphen1 (changes requested):
Several changes were requested, check tests before adding changes.

PR #640, opened 2023-07-28
py-person-class

stamaksim:


vsmutok (changes requested):
remove file from GitHub
image

stamaksim:
file deleted

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+        else:
+            person = Person(person_name, person_data["age"])
+        if "wife" in person_data and person_data["wife"] in Person.people:
+            person.wife = Person.people[person_data["wife"]]
+        elif "wife" in person_data and person_data["wife"] is not None:
+            person.wife = None
+        if "husband" in person_data \
+                and person_data["husband"] in Person.people:
+            person.husband = Person.people[person_data["husband"]]
+        person_list.append(person)
```
ArsenPidhoretskyi, line 25: Better to split logic into two loops. In first loop create all person. In second loop set all wife/husband

app/main.py
```diff
@@ -1,8 +1,26 @@
...
-    # write your code here
-    pass
+    person_list = []
+    for person_data in people:
+        person_name = person_data["name"]
+        if person_name in Person.people:
+            person = Person.people[person_name]
+        else:
+            person = Person(person_name, person_data["age"])
+        if "wife" in person_data and person_data["wife"] in Person.people:
```
ArsenPidhoretskyi, line 18: You can use .get instead of two conditions:
if Person.people.get(person_data.get("wife"))

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: not fixed

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        if person_name not in Person.people:
+            Person(person_data["name"], person_data["age"])
+
+    for person_data in people:
+        person = Person.people[person_data["name"]]
+
+        if "wife" in person_data and person_data["wife"] in Person.people:
+            person.wife = Person.people[person_data["wife"]]
+        elif "wife" in person_data and person_data["wife"] is not None:
+            person.wife = None
```
ArsenPidhoretskyi, line 22: this if statement is not needed

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+            Person(person_data["name"], person_data["age"])
+
+    for person_data in people:
+        person = Person.people[person_data["name"]]
+
+        if Person.people.get(person_data.get("wife")):
+            person.wife = Person.people[person_data["wife"]]
+
+        if "husband" in person_data\
+                and person_data["husband"] in Person.people:
```
ArsenPidhoretskyi, line 23: Do the same as with wife

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+            person.husband = Person.people[person_data["husband"]]
+
+    return list(Person.people.values())
```
ArsenPidhoretskyi, line 26: Also you need to return only just created people, not all people

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #642, opened 2023-07-30
Solution

liliia-kostenko:


app/main.py
```diff
@@ -1,8 +1,33 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
```
ArsenPidhoretskyi, line 7: You can use self. instead of Person.
```suggestion
        self.people[name] = self
```

app/main.py
```diff
@@ -1,8 +1,33 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons_list = []
```
ArsenPidhoretskyi, line 11: persons is not correct. Person - one, People - Many
Use people_objs

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #646, opened 2023-08-02
Implemented class Person and function create_person_list

sberdianskyi:


app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
```
ArsenPidhoretskyi, line 7: FYI 
You can use self.people[name] = self

app/main.py
```diff
@@ -1,8 +1,21 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for friend in people:
+        person = Person(friend["name"], friend["age"])
+        person_list.append(person)
+    for friend in people:
+        if "wife" in friend and friend["wife"] is not None:
```
ArsenPidhoretskyi, line 16: Use .get instead of and statement
friend.get("wife") is not None

app/main.py
sberdianskyi: changes made to the code in the last commit

app/main.py
sberdianskyi: changes made to the code in the last commit

ArsenPidhoretskyi (approved):
Well done!

PR #649, opened 2023-08-03
Solution

nyanplague:


RPodolets (approved):
Good solution :)

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+
+        Person.people[name] = self
+
+    @classmethod
+    def create_instances(cls, list_of_people: list[dict]) -> None:
```
Kroll410, line 12: This breaks the SOLID principles, but you'll study it after. For now you can check [this](https://en.wikipedia.org/wiki/SOLID)  and why `create_instances` is a bad method for `Person` classs

Kroll410 (approved):

PR #650, opened 2023-08-04
Solution

olgierrd:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
```
Polyakiv-Andrey, line 10: more detail type annotations

Polyakiv-Andrey (changes requested):

app/main.py
olgierrd: what do you mean by "more detailed"?

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+    people_dict = {}
+    for person in people:
+        name = person["name"]
+        age = person["age"]
+        person_instance = Person(name, age)
+        people_dict[name] = person_instance
+
+    for person in people:
+        name = person["name"]
+        if "wife" in person and person["wife"] is not None:
```
pavlejviki, line 20: You can use .get() method instead of two conditions.

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list:
+    people_dict = {}
```
pavlejviki, line 11: The Person.people dictionary is being updated every time a Person instance is created. This means that you can directly access the Person.people dictionary inside your create_person_list function without needing the local people_dict.

So, you can remove the people_dict and use Person.people instead, simplifying your code.

pavlejviki (changes requested):

pavlejviki (approved):
Good job!

PR #651, opened 2023-08-05
task completed, but an error occurs in the test that is not related t…

DaniilHinzburh:
…o the code, but the code is probably correct

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,33 @@
+from __future__ import annotations
+
+
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.add_object(self)
+
+    @classmethod
+    def add_object(cls, obj: Person) -> None:
+        cls.people[obj.name] = obj
 
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+def create_person_list(people: list[dict]) -> list[Person]:
+    result = []
+
+    for pers in people:
+        person = Person(name=pers["name"], age=pers["age"])
+        if "wife" in pers and pers["wife"] is not None:
+            for i in Person.people.values():
+                if i.name == pers["wife"]:
+                    person.wife = i
+                    i.husband = person
+        elif "husband" in pers and pers["husband"] is not None:
+            for i in Person.people.values():
+                if i.name == pers["husband"]:
+                    person.husband = i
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,33 @@
+from __future__ import annotations
```
vsmutok, line 1: try to remove it, maybe it's help you

vsmutok (changes requested):
Plese fix tests

DaniilHinzburh:
if I remove it, then there will be an error in this part of the code:

@classmethod
def add_object(cls, obj: Person) -> None:
    cls.people[obj.name] = obj

for some reason the tests can't work with everyone "from ##### import #####"

I fixed the error by specifying obj as an object

however, in the future it may interfere with other tasks, perhaps a problem
with the tests or I have with my compiler

сб, 5 авг. 2023 г. в 20:56, Vasyl Smutok ***@***.***>:

vasylhnatiuk (changes requested):
This code works, but it's really hard to read. Don't use `i` like a name for Vairible. Use only two loops. first for creating Person instances and another for setting husband and wife.

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+    def add_object(cls, obj: object) -> None:
+        cls.people[obj.name] = obj
+
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    result = []
+
+    for pers in people:
+        person = Person(name=pers["name"], age=pers["age"])
+        result.append(person)
```
vsmutok, line 19: use list comprehension

app/main.py
```diff
@@ -1,8 +1,36 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.add_object(self)
```
vsmutok, line 7:
```suggestion
        self.people[name] = self
```

app/main.py
```diff
@@ -1,8 +1,36 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    @classmethod
+    def add_object(cls, obj: object) -> None:
+        cls.people[obj.name] = obj
+
```
vsmutok, line 12: Redundant here

vsmutok (changes requested):

vsmutok (approved):
Good Job 👍

PR #652, opened 2023-08-05
class Person and func create_person_list implemented, tests passed

maxkatkalov:
Solution implemented, please check!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int, **kwargs) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,27 @@
...
-    pass
+
+def create_person_list(people: list[dict]) -> list:
+    res_person_instances = [
+        Person(person["name"], person["age"])
+        for person in people
+    ]
+
+    for person in people:
+        if person.get("wife") is not None:
```
Kroll410, line 21: Just  person.get("wife")

Kroll410 (changes requested):

app/main.py
maxkatkalov: done

CaCtUs1k (approved):

PR #653, opened 2023-08-06
Solution for py-person-class

Barsh4ec:


KristianYe (approved):

app/main.py
```diff
@@ -1,8 +1,28 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        person_name = person["name"]
+        person_age = person["age"]
+        person_list.append(Person(person_name, person_age))
+
+    for index, person in enumerate(person_list):
```
Kroll410, line 17: Why do we use enumerate here?

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        if people[index].get("wife") is not None:
+            wife = [item
+                    for item in person_list
```
Kroll410, line 20: Too complicated, try different looping from the 17th line of code

Kroll410 (changes requested):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = []
+    for person in people:
+        human = Person(person["name"], person["age"])
+        if person.get("wife"):
+            for human2 in people:
+                if person["wife"] == human2["name"]:
+                    person_wife = Person(human2["name"], human2["age"])
```
IvanRamyk, line 17: I believe you will create x2 instances of the Person class with this approach. Try to do this task in two steps:

create all people without partners
fill partners with existing instances (use Person.people list to find a proper partner)

IvanRamyk (approved):

PR #654, opened 2023-08-07
'Solution'

lgSilay:


innyshka (approved):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_people_list = []
+
+    for person in people:
+        new_people_list.append(Person(person["name"], person["age"]))
+
+    for person in people:
+        if person.get("wife") is not None:
```
Kroll410, line 17: Simply person.get("wife")

Kroll410 (changes requested):

app/main.py
lgSilay: Your right!) Already corrected)

PR #673, opened 2023-08-08
'Solution'

ochernous:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
+def create_person_list(people: list[dict]) -> list[Person]:
+    person_list = []
+
+    for man in people:
+        person = Person(man["name"], man["age"])
+
+        if "wife" in man and man["wife"] is not None:
+            for man2 in people:
+                if man["wife"] == man2["name"]:
+                    wife = Person(man2["name"], man2["age"])
```
IvanRamyk, line 19: I believe you will create x2 instances of the Person class with this approach. Try to do this task in two steps:

1. create all people without partners
2ю fill partners with existing instances (use Person.people list to find a proper partner)

IvanRamyk (changes requested):

app/main.py
ochernous: I had some doubts about this part - but tests passed, so I thought everything  was ok

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    person_list = []
+
+    for man in people:
+        person = Person(man["name"], man["age"])
+        person_list.append(person)
+
+    for man in people:
+        if "wife" in man and man["wife"] is not None:
```
IvanRamyk, line 18: use `get` to simplify the condition

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+            husband = Person.people[man["name"]]
```
IvanRamyk, line 19: I believe this logic should be a part of the another branch where "husband" is in `man` dict

app/main.py
```diff
@@ -1,8 +1,24 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    person_list = []
+
+    for man in people:
+        person = Person(man["name"], man["age"])
```
IvanRamyk, line 14: extra variable
also consider rewriting it using list comprehension

IvanRamyk (changes requested):

app/main.py
ochernous: I thought that using "for...in" in this case improves code readability )

IvanRamyk (approved):
well done!

PR #675, opened 2023-08-09
Develop

IvDyriv:


app/main.py
```diff
@@ -1,8 +1,32 @@
...
+    result_list: list[Person] = []
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person_instance = Person(name, age)
+        result_list.append(person_instance)
+
+    for person_data in people:
+        name = person_data["name"]
+        spouse_name = person_data.get("wife") or person_data.get("husband")
```
YukihiroSM, line 23: This gives you True/False, not the exact spouse name
Are you sure, that spouse_instance would be found?

YukihiroSM (changes requested):

app/main.py
IvDyriv: > This gives you True/False, not the exact spouse name Are you sure, that spouse_instance would be found?

I'm sorry, but I don't understand what exactly the error is, all tests work.the code assumes that the spouse_instance will be found based on the spouse's name retrieved from the person_data dictionary.

app/main.py
```diff
@@ -1,8 +1,32 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result_list: list[Person] = []
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person_instance = Person(name, age)
+        result_list.append(person_instance)
```
Bohdan-Salamakha, line 19: use list comprehension here

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #681, opened 2023-08-11
Solution

o-davydova:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for index, person in enumerate(people):
+        partner = person_list[index]
+
+        if "wife" in person and person["wife"] is not None:
```
pavlejviki, line 16: you can use get() method instead of two conditions

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+            wife = Person.people[person["wife"]]
+            partner.wife = wife
+        elif "husband" in person and person["husband"] is not None:
```
pavlejviki, line 19: you can use get() method instead of two conditions

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #687, opened 2023-08-15
Solution

prochigor:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_data in people:
+        person_list.append(Person(person_data["name"], person_data["age"]))
+
+    for i in people:
```
LiudmylaKulinchenko, line 15: Use descriptive name instead of `i`

app/main.py
```diff
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```

app/main.py
```diff
@@ -1,8 +1,22 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_data in people:
+        person_list.append(Person(person_data["name"], person_data["age"]))
```

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+    for i in people:
+        person = Person.people[i["name"]]
+        if "wife" in i and i["wife"]:
```

LiudmylaKulinchenko (changes requested):

Naz-iv (changes requested):
please implement changes requested by mentor

app/main.py
prochigor: Done

app/main.py
prochigor: good idea, done

app/main.py
prochigor: I see

app/main.py
prochigor: Done

fsocie7y (approved):
well done 🚀

PR #693, opened 2023-08-18
Attempt2

sashabryl:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list:
+    for person in people:
+        person_info = list(person.values())
+        Person(person_info[0], person_info[1])
+    for person in people:
+        if "wife" in person:
+            if person["wife"] is not None:
```
pavlejviki, line 16: use get() method instead of nested conditions

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+                person_link = list(person.values())[0]
+                husband = Person.people[person_link]
+                husband.wife = Person.people[person["wife"]]
+        if "husband" in person:
+            if person["husband"] is not None:
```
pavlejviki, line 21: use get() method instead of nested conditions

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        Person.people[self.name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list:
+    for person in people:
+        person_info = list(person.values())
+        Person(person_info[0], person_info[1])
```
pavlejviki, line 13: you can simply access the value of name and age by its key, there is no need to convert it to a list.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #694, opened 2023-08-20
Implemented person_class

osharlai:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_instances = []
+
+    for person in people:
+        list_of_instances.append(Person(person["name"], person["age"]))
```
pavlejviki, line 14: Consider using list comprehension

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```
pavlejviki, line 17: use get() method instead of two conditions

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+            Person.people[person["name"]].wife \
+                = Person.people[person["wife"]]
```
pavlejviki, line 19:
```suggestion
Person.people[person["name"]].wife = (
    Person.people[person["wife"]]
)

```

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        if "husband" in person and person["husband"] is not None:
+            Person.people[person["name"]].husband \
+                = Person.people[person["husband"]]
```
pavlejviki, line 22: use parenthesis to avoid theusage of \

pavlejviki (changes requested):

app/main.py
osharlai: Done

app/main.py
osharlai: good advice

app/main.py
osharlai: fixed

PR #697, opened 2023-08-21
Solution

Daniil-Pankieiev:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for person_info in people:
+        name = person_info["name"]
+        age = person_info["age"]
+        person = Person(name, age)
```
LiudmylaKulinchenko, line 15: Use list comprehentsion to create list of Person instances

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+
+        if "wife" in person_info and person_info["wife"] is not None:
```

LiudmylaKulinchenko (changes requested):

Familenko (approved):

PR #699, opened 2023-08-22
Person_solution

Filllick:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    list_of_people = []
```
ArsenPidhoretskyi, line 12: rename list_of_people to people_objs

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+
+    for data in people:
+        person_inst = Person(data["name"], data["age"])
+        list_of_people.append(person_inst)
+
+    for data in people:
+        for names in Person.people.keys():
```
ArsenPidhoretskyi, line 19: This loop is redundant

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+            if "wife" in data and data["wife"] == names:
```
ArsenPidhoretskyi, line 20: Use data.get("wife") instead of two conditions

ArsenPidhoretskyi (changes requested):

app/main.py
Filllick: Renamed  list_of_people, used .get to look through keys, replaced nested loop

Filllick:
Renamed list_of_people, used .get to look through keys, replaced nested loop

PR #700, opened 2023-08-24
solution

SerhiiBudko:


app/main.py
```diff
@@ -1,8 +1,32 @@
...
+        age = person_info["age"]
+        person_instance = Person(name, age)
+        person_instances[name] = person_instance
+
+    for person_info in people:
+        person_instance = person_instances[person_info["name"]]
+
+        if "wife" in person_info:
+            wife_name = person_info["wife"]
+            if wife_name is not None:
```
ArsenPidhoretskyi, line 24: Just use: person_info.get("wife") is not None
No need to use two if conditions
The same with husband

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_instances = {}
```
ArsenPidhoretskyi, line 11: This can be list also

ArsenPidhoretskyi (changes requested):

pavlejviki (approved):
Well done!

PR #702, opened 2023-08-27
Solution

Graniecki:


app/main.py
```diff
@@ -1,8 +1,47 @@
...
+
+    for person in people:
+        person_instance = None
+
+        if person["name"] in Person.people.keys():
+            person_instance = Person.people[person["name"]]
+        else:
+            person_instance = Person(person["name"], person["age"])
+
+        persons_list.append(person_instance)
```
Bohdan-Salamakha, line 21: use list comprehension here

app/main.py
```diff
@@ -1,8 +1,47 @@
...
+                person_instance.husband = Person.people[person["husband"]]
+            else:
+                husband = [
+                    human
+                    for human in people
+                    if human["name"] == person["husband"]
+                ][0]
+                person_instance.husband = Person(
+                    husband["name"], husband["age"]
+                )
```
Bohdan-Salamakha, line 45: u don't need to do all this in the first for cycle, create the second one

Bohdan-Salamakha (changes requested):

PR #703, opened 2023-08-27
Develop

YaroslavYarovoy:


app/main.py
```diff
@@ -1,8 +1,34 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = []
+    for human in people:
+        name = human["name"]
+        age = human["age"]
+        pers = Person(name, age)
+
+        if "wife" in human:
```
pavlejviki, line 19: better use get() method to check if the key exists and if its not None

app/main.py
```diff
@@ -1,8 +1,34 @@
...
+            wife_name = human["wife"]
+            wife = Person.people.get(wife_name)
+            if wife:
+                pers.wife = wife
+                wife.husband = pers
+
+        if "husband" in human:
```
pavlejviki, line 26: better use get() method to check if the key exists and if its not None

app/main.py
```diff
@@ -1,8 +1,34 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = []
+    for human in people:
+        name = human["name"]
+        age = human["age"]
+        pers = Person(name, age)
```
pavlejviki, line 17: pers is not descriptive variable name

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #729, opened 2023-09-11
implemented py-person-class task

HlovaOleh:


app/main.py
```diff
@@ -1,8 +1,33 @@
...
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
```
Bohdan-Salamakha, line 9: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

Bohdan-Salamakha (changes requested):

olena-hul (changes requested):

olena-hul (approved):
Good job!

PR #742, opened 2023-09-21
'Solution'

Prorochenkko:


app/main.py
```diff
@@ -1,8 +1,31 @@
...
+        name = person_data["name"]
+        age = person_data["age"]
+
+        person_list.append(Person(name, age))
+
+    for person_data in people:
+        name = person_data["name"]
+        person = Person.people[name]
+
+        if "wife" in person_data and person_data["wife"]:
```
ArsenPidhoretskyi, line 26: Instead of 'and' use .get for dict

ArsenPidhoretskyi (approved):
Well done!

PR #748, opened 2023-09-26
'Solution'

Illya-Maznitskiy:


app/main.py
```diff
@@ -1,8 +1,36 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.wife = None
+        self.husband = None
```
pavlejviki, line 8: The instance should not have wife/husband attributes by default

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        name = data["name"]
+        age = data["age"]
+        wife = data.get("wife")
+        husband = data.get("husband")
+
+        person = Person(name, age)
+        person_list.append(person)
+
+        if wife is not None:
+            if wife in Person.people:
```
pavlejviki, line 24: you can use 'and' instead of nested if

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+                person.wife = Person.people[wife]
+                Person.people[wife].husband = person
+
+        if husband is not None:
+            if husband in Person.people:
+                person.husband = Person.people[husband]
+                Person.people[husband].wife = person
+
+        if person.wife is None:
+            del person.wife
```
pavlejviki, line 34: after you remove redundant  attributes from init , you won't need this block of code

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+    person_list = []
+    for data in people:
+        name = data["name"]
+        age = data["age"]
+        wife = data.get("wife")
+
+        person = Person(name, age)
+        person_list.append(person)
+
+        if wife is not None and wife in Person.people:
```
viktoria-rybenchuk, line 20: when you want to check if a variable is not None, you can use a simple if wife

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #749, opened 2023-09-27
'Solution'

lylakivan:


app/main.py
```diff
@@ -1,8 +1,33 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result_list = []
+
+    for person_list in people:
+        name = person_list["name"]
+        wife_or_husband = list(person_list.keys())[2]
```
pavlejviki, line 15: you can 
```suggestion
        spouse = list(person_list.keys())[2]
```

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+            Person.people[name] = Person(name, person_list["age"])
+        if (person_list[wife_or_husband] is not None
+                and person_list[wife_or_husband] in Person.people):
+            setattr(Person.people[name],
+                    wife_or_husband,
+                    Person.people[person_list[wife_or_husband]])
+        elif person_list[wife_or_husband] is not None:
+            setattr(Person.people[name],
+                    wife_or_husband,
+                    Person(person_list[wife_or_husband], 0))
```
pavlejviki, line 30: why are you setting the age as 0?

you should consider a different approach for solving this task.

first you can create all the instances and than in a separate loop set the relations without checking the existence of the instance in Person.people

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #752, opened 2023-09-29
solution for person_class

wataschi:
Hi
Is there any way to simplify the check cycles for "husband" and "wife"?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int, *args) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+def create_person_list(people_list: list) -> list:
+
+    persons = [Person(per["name"], per["age"]) for per in people_list]
+
+    for people, value in zip(persons, people_list):
+        if "wife" in value and value["wife"]:
+            for count in Person.people:
+                if count == value["wife"]:
+                    people.wife = Person.people[count]
+        if "husband" in value and value["husband"]:
+            for count in Person.people:
+                if count == value["husband"]:
+                    people.husband = Person.people[count]
+    return persons


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
+
+    persons = [Person(per["name"], per["age"]) for per in people_list]
```
pavlejviki, line 12: per is not a descriptive variable name

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+
+    for people, value in zip(persons, people_list):
+        if "wife" in value and value["wife"]:
```
pavlejviki, line 15: use .get() instead of two conditions

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+            for count in Person.people:
```
pavlejviki, line 16: why do you need this loop? you can get the necessary instance from the dictionary by its key Person.people[value["wife"]]

pavlejviki (changes requested):

app/main.py
wataschi: Corrected on all counts :)

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    def __init__(self, name: str, age: int, *args) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
```
Bohdan-Salamakha, line 10: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+
+    persons = [Person(temp_person["name"],
+                      temp_person["age"])
+               for temp_person in people_list
+               ]
```
Bohdan-Salamakha, line 15:
```suggestion
    persons = [
               Person(temp_person["name"], temp_person["age"])
               for temp_person in people_list
               ]
```

Bohdan-Salamakha (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #754, opened 2023-09-30
Solution

Naz-iv:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
+
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
+
+    def add_husband_wife_if_merried(self, person: dict) -> None:
+        if person.get("wife") is not None:
```
prochigor, line 11: It is not necessary to write "is not None" in this case.

prochigor (commented):
In the README.md file, only one function and class Person do not have methods, only__init__, so it isn't like in documentation, but a possible decision.

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+            self.wife = Person.people.get(person.get("wife"))
+
+        if person.get("husband") is not None:
+            self.husband = Person.people.get(person.get("husband"))
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
```
prochigor, line 19: two identical cycles, maybe they can be combined

prochigor (changes requested):

app/main.py
Naz-iv: noted, thanks

app/main.py
Naz-iv: they are not identical, first is to initialize instances of a class Person, second to link wife/husband reference to a instance of a class if he/she has wife/husband.

Naz-iv:
I decided to make it be a method since it's changing state of instance of the class person. So in case my person get's married or divorced, instance of a class can be updated

app/main.py
prochigor: Yes, I see, if do that in one cycle, can be error

fsocie7y (approved):
so intresting solution 💯

PR #756, opened 2023-10-01
Solution

frezworx:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
+    for data_person in peoples:
+        Person(
+            name=data_person["name"],
+            age=data_person["age"]
+        )
+
+    for data_person in peoples:
+        person = Person.people[data_person["name"]]
+
+        if "wife" in data_person and data_person["wife"]:
```
Dimosphen1, line 24: You could write this check like
```suggestion
        if data_person.get("wife"):
```
It would automatically check that there is a `"wife"` key in the `data_person` dict without raising a `KeyError` and evaluate the boolean type of the value in case it is present.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #759, opened 2023-10-02
Solution

DenVia:


app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for i in range(len(people)):
+        person_list.append(Person(people[i]["name"], people[i]["age"]))
```
LiudmylaKulinchenko, line 14: Rewrite in one line using list comprehension

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+
+    for i in range(len(people)):
+        if "wife" in people[i] and people[i]["wife"]:
```

app/main.py
```diff
@@ -1,8 +1,29 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for i in range(len(people)):
+        person_list.append(Person(people[i]["name"], people[i]["age"]))
+
+    for i in range(len(people)):
```
LiudmylaKulinchenko, line 16: U can iterate by person here, not by index. Don't forget to use descriptive name

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        if "wife" in people[i] and people[i]["wife"]:
+            person = Person.people[people[i]["name"]]
+            wife = Person.people.get(people[i]["wife"])
+            if wife:
+                person.wife = wife
+
+        if "husband" in people[i] and people[i]["husband"]:
+            person = Person.people[people[i]["name"]]
+            husband = Person.people.get(people[i]["husband"])
+            if husband:
```
LiudmylaKulinchenko, line 26: husband anyway is in the Person.people, you checked it in 23 line, so no need to check here

app/main.py
LiudmylaKulinchenko: Mind not to iterate by indexes also here

LiudmylaKulinchenko (changes requested):

app/main.py
DenVia: I must try to avoid all iterations by indexes? I kinda like them, since it's easier to navigate through the iteration for me

LiudmylaKulinchenko: You can avoid them here to write less code, so you won't need to get a person by index and then name. Without indexes here you'll already have a person and then just get a name

LiudmylaKulinchenko (approved):
Good job 🤗

PR #762, opened 2023-10-02
Solution

mykolamateichuk:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person in people:
+        person_list.append(Person(person["name"], person["age"]))
+
```
petrykivd, line 16: You can use here a list comprehension, it's look here more consice.

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+    for index, person in enumerate(people):
+        if "wife" in person and person["wife"]:
+            person_list[index].wife = Person.people[person["wife"]]
+        if "husband" in person and person["husband"]:
+            person_list[index].husband = Person.people[person["husband"]]
```
petrykivd, line 21: It is bad practice to refer to the dictionary by key. Here you check whether it exists, but do you need it? You can use .get(), so you will increase the readability of the code and not get an error if the key does not exist.

petrykivd (changes requested):

RabenkoYevhenii (commented):
Im agree with the petrykivd`s comments

RabenkoYevhenii (approved):
It is better now. I think it is well-optimized and has a correct structure.

fsocie7y (approved):
well done ^_^

PR #773, opened 2023-10-03
Solution

MiskoRuslan:
🤔

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
+
+    @classmethod
+    def couple_create(cls, name: str, couple_name: str, is_male: bool) -> None:
+        if couple_name in cls.people and is_male:
+            cls.people[name].wife = cls.people[couple_name]
+        else:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
```
roffi37: missing return type annotationation

roffi37 (approved):

DenPrislipskyi (approved):

fedorov-s-od (approved):
Nice solution! I used more risky option with "Person.people[name] = self" in line 7. Your option looks better. Noticed! ))

app/main.py
```diff
@@ -1,8 +1,25 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
fsocie7y, line 7: In this case, it is better to refer to the class attribute through the class name, and not like this 
`self.__class__.people[name] = self`

everything else is good)

fsocie7y (changes requested):

fsocie7y (approved):
GJ 🚀

PR #774, opened 2023-10-03
Solution

roffi37:


Shatrovskyi (approved):

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    for index, person in enumerate(people):
+        wife = person.get("wife")
+        if wife is not None:
```
fsocie7y, line 14: `Is not None` is redundant here, since in your case if statement will be True if key exists

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+            person_list[index].wife = Person.people.get(wife)
+        husband = person.get("husband")
+        if husband is not None:
```
fsocie7y, line 17: same

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #778, opened 2023-10-04
Solution

Shatrovskyi:


Shatrovskyi:
To be honest, I think my implementation of the task is not good. Can someone help to implement this task in a more convenient way?

Shatrovskyi:
Now It may be better)

app/main.py
```diff
@@ -1,8 +1,28 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
```
Andrii-Silenko, line 7:
```suggestion
        self.pople[name] = self
```
maybe it will be better

Andrii-Silenko (approved):

app/main.py
fsocie7y: no, here u must use class atribute not object of class

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        person = Person(name, age)
+
+        person_list.append(person)
+
+    for index, person in enumerate(people_data):
+        wife = person.get("wife")
+        husband = person.get("husband")
+        if wife is not None:
+            person_list[index].wife = Person.people.get(wife)
+        if husband is not None:
```
fsocie7y, line 25: same

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_data: dict) -> list[Person]:
+    person_list = []
+
+    for person_item in people_data:
```
fsocie7y, line 13: Try to rewrite in one line. Hint: list comprehension 🤫

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        name = person_item["name"]
+        age = person_item["age"]
+        person = Person(name, age)
+
+        person_list.append(person)
+
+    for index, person in enumerate(people_data):
+        wife = person.get("wife")
+        husband = person.get("husband")
+        if wife is not None:
```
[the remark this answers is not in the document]
Shatrovskyi: Got it!

fsocie7y (approved):
GJ 👍

PR #780, opened 2023-10-04
Solution

Andrii-Silenko:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    pass
+
+def create_person_list(people_to_create: list) -> list:
+    person_list = [Person(name=person["name"], age=person["age"])
+                   for person in people_to_create]
+
+    for person in people_to_create:
+        if "wife" in person.keys() and person["wife"] is not None:
+            Person.people[person["name"]].wife = \
+                Person.people[person["wife"]]
```
LiudmylaKulinchenko, line 17: Don't use `\`, create additional variable. Create it for `Person.people[person["name"]]`, for example

app/main.py
```diff
@@ -1,8 +1,22 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_to_create: list) -> list:
+    person_list = [Person(name=person["name"], age=person["age"])
+                   for person in people_to_create]
+
+    for person in people_to_create:
+        if "wife" in person.keys() and person["wife"] is not None:
```
LiudmylaKulinchenko, line 15: Use just dict.get() method in the condition, instead of this long one

app/main.py
```diff
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 😸🎈

PR #784, opened 2023-10-05
'Solution'

Tituha:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = []
+    for human in people:
+        person = Person(human["name"], human["age"])
+        if "wife" in human and human["wife"] is not None:
```
sakhaline, line 14: You can write just "human['wife']" instead of "human["wife"] is not None" because bool(None) is False.

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+            for human2 in people:
+                if human["wife"] == human2["name"]:
+                    person_wife = Person(human2["name"], human2["age"])
+                    person_wife.husband = person
+                    person.wife = person_wife
+        persons.append(person)
+    return persons
```
sakhaline, line 21: In general I guess it would be nice to use list comprehension instead of nested loops.

Oxbay (approved):
Good job!

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = []
```
LiudmylaKulinchenko, line 11: Rewrite list creation in one line

app/main.py
LiudmylaKulinchenko, line 14: Use just dict.get() method in the condition, instead of this long one

LiudmylaKulinchenko (changes requested):

PR #787, opened 2023-10-07
Solution

No4-KilL:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons_ls = []
+
+    for body in people:
+        person = Person(body["name"], body["age"])
+        persons_ls.append(person)
```
viktoria-rybenchuk, line 15: use a list comprehension

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+
+    for i in range(len(people)):
```
viktoria-rybenchuk, line 17: why do you use `range` here?

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+
+        if "wife" in people[i] and people[i]["wife"] is not None:
```
viktoria-rybenchuk, line 19: don't check explicitly for None

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (changes requested):

PR #793, opened 2023-10-09
Solution

OleksandrMV:


app/main.py
```diff
@@ -1,8 +1,17 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people.update({self.name: self})
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons_list = [Person(one["name"], one["age"]) for one in people]
```
ArsenPidhoretskyi, line 11:
```suggestion
    persons_list = [Person(person["name"], person["age"]) for person in people]
```
one - is bad variable name

app/main.py
```diff
@@ -1,8 +1,17 @@
...
+    for key, value in enumerate(people):
```
ArsenPidhoretskyi, line 12: key is bad variable name
also the same with value

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #796, opened 2023-10-10
Solution

paashkovaaa:


app/main.py
```diff
@@ -1,8 +1,35 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_info in people:
+        name = person_info["name"]
+        age = person_info["age"]
+        person = Person(name, age)
```
EdAlekseiev, line 16: You must create all Persons first, and after this assign wives and husbands

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,8 +1,34 @@
...
+            wife = Person.people.get(wife_name)
+            if wife:
+                person.wife = wife
+                wife.husband = person
+
+        if husband_name is not None:
+            husband = Person.people.get(husband_name)
+            if husband:
+                person.husband = husband
+                husband.wife = person
```
YukihiroSM, line 32: This should be in separate loop after creating all persons, because husband or wife person might be not yet created at this point.

app/main.py
```diff
@@ -1,8 +1,34 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_info in people:
+        name = person_info["name"]
+        age = person_info["age"]
+        person = Person(name, age)
+        person_list.append(person)
```
YukihiroSM, line 17: What about using list comprehension?

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #800, opened 2023-10-12
Solution

maksimpro001:


app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    pep = [Person(p["name"], p["age"]) for p in people]
+    for i, pop in enumerate(people):
```
ArsenPidhoretskyi, line 12: pop is not good variable name

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    pep = [Person(p["name"], p["age"]) for p in people]
```
ArsenPidhoretskyi, line 11: p, pep - bad variable names

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    for i, pop in enumerate(people):
+        try:
+            if pop["wife"] is not None:
+                pep[i].wife = Person.people[pop["wife"]]
+        except KeyError:
```
ArsenPidhoretskyi, line 16: why do you need to use try except? It is redundant, use correct if statements

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
good job!

PR #804, opened 2023-10-13
Add solution

CrazyHilly:


app/main.py
```diff
@@ -1,8 +1,31 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_dict = Person.people
+    for person in people:
+        person_name = person["name"]
+        person_age = person["age"]
+        spouse = "wife" if "wife" in list(person.keys()) else "husband"
+        spouse_name = person[spouse]
```
YukihiroSM, line 16: you don't need to do this as you already have this data in people list (which is on function entry)

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+
+        if spouse_name:
+            setattr(people_dict[person_name], spouse, spouse_name)
+
+    for person, obj in people_dict.items():
+        spouse = [s for s in obj.__dict__ if s in ["wife", "husband"]]
+        if not spouse:
+            continue
+        spouse_name = obj.__dict__[spouse[0]]
+        setattr(people_dict[person], spouse[0], people_dict[spouse_name])
```
YukihiroSM, line 29: What about iterating over people in the input list, find person in Person.person by key, find wife or husband by the key as well (because all persons are in People.people

YukihiroSM (changes requested):

app/main.py
CrazyHilly: Thanks for the suggestion. 
To minimize the number of iterations on this step, I also decided to create a separate dict of {husband: wife} pairs beforehand to have easy access to info about all spouses. I then iterate through it when updating spouse info in "people" dictionary.

EdAlekseiev (approved):
Good job

PR #808, opened 2023-10-17
Solution

IvanMozhar:


app/main.py
```diff
@@ -1,8 +1,35 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    @classmethod
+    def __repr__(cls) -> str:
+        return f"{cls.people}"
+
+
+def create_person_list(people: dict) -> list:
```
Dimosphen1, line 14: `people` argument expects not dict, but a list of dicts. Besides that, you could specify that you return a list of Person instances in return type annotation.

app/main.py
```diff
@@ -1,8 +1,35 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    @classmethod
+    def __repr__(cls) -> str:
+        return f"{cls.people}"
```
Dimosphen1, line 11: There is no need for `__repr__` in this class, delete it.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #810, opened 2023-10-18
Solution

okien1:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = []
+    for person in people:
+        persons.append(Person(person["name"], person["age"]))
```
pavlejviki, line 13: you can use list comprehension instead

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+    for i, person in enumerate(people):
```
pavlejviki, line 14: i is not a descriptive variable name

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        if "wife" in person and person["wife"] is not None:
```
pavlejviki, line 15: use get() method instead of two conditions

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #814, opened 2023-10-21
Solution

AnyoneClown:


app/main.py
```diff
@@ -1,8 +1,34 @@
...
-    pass
+
+def create_person_list(people: Person) -> list:
+    person_instances = []
+
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+
+        person = Person(name, age)
```
viktoria-rybenchuk, line 17: you can create Person instances using list comprehension and then set the wife/husband attribute separately using another loop

viktoria-rybenchuk (changes requested):

vsmutok (approved):
looks Good 👍

PR #815, opened 2023-10-22
solution

Andrey1104:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    friends = [Person(someone["name"], someone["age"]) for someone in people]
```
pavlejviki, line 12: choose better variable names.  friends and someone are  not appropriate.

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    for friend in friends:
```
pavlejviki, line 13: you can simplify the code by first creating all the Person objects, and then, in a subsequent loop, assigning the husbands and wives. At this stage, the husband or wife Person object will have already been created, facilitating this assignment.

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        for person in people:
+            if "wife" in person \
```
pavlejviki, line 15: use () instead of \

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
Dimosphen1, line 8: Call `people` not from `self` but from the `Person` class, it is a class attribute.

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+    humans = [Person(person["name"], person["age"]) for person in people]
+    for human in humans:
+        for person in people:
+            if (
+                "wife" in person
+                and person["wife"] is not None
+                and human.name == person["name"]
+            ):
+                human.wife = Person.people[person["wife"]]
+            elif "husband" in person and person["husband"] is not None:
```
Dimosphen1, line 21: You could write this check in one line, e.g.:
```suggestion
            elif person.get("husband"):
```
You could make your previous check shorter as well with the same approach.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #816, opened 2023-10-22
Solution

Dikey945:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+    for person in people:
+        if person["name"] not in Person.people:
+            Person(person["name"], person["age"])
+
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"] is not None:
+            Person.people[person["name"]].husband \
+                = Person.people[person["husband"]]
```
LiudmylaKulinchenko, line 20: Better to use parentheses instead of `\` to split long lines

app/main.py
```diff
@@ -1,8 +1,22 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        if person["name"] not in Person.people:
+            Person(person["name"], person["age"])
+
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"] is not None:
+            Person.people[person["name"]].husband \
+                = Person.people[person["husband"]]
+
+    return list(Person.people.values())
```
LiudmylaKulinchenko, line 22: You can create the list from the start of the function

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        if person["name"] not in Person.people:
+            Person(person["name"], person["age"])
```
LiudmylaKulinchenko, line 13: Use list comprehension here

LiudmylaKulinchenko (changes requested):

PR #818, opened 2023-10-23
Created the Person class and implemented the create_person_list method.

lyaskov:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
+        self.age = age
+        self.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        if person["name"] not in Person.people.keys():
+            Person(person["name"], person["age"])
```
EdAlekseiev, line 13: You can use list comprehension here

EdAlekseiev (changes requested):

app/main.py
lyaskov: I can write here...
[Person(person["name"], person["age"]) for person in people if person["name"] not in Person.people.keys()]

But in my opinion, the code of this line does not look more clear and attractive. And created list will not used in the future.
Or I'm wrong?

YukihiroSM: You don't need to check if the name is in Person.people. Also, you need to return this list instead of transforming Person.people.values() into list. So list comprehension is quite nice way to do this.

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #819, opened 2023-10-24
Develop

gesiod:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
-    # write your code here
-    pass
+    person_list = []
+
+    for people_dict in people:
+        person = Person(people_dict["name"], people_dict["age"])
+        person_list.append(person)
+
+    for ind, spouses in enumerate(people):
+        if "wife" in spouses and spouses["wife"] is not None:
```
Polyakiv-Andrey, line 18: is not None should be omitted, it will work without it

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+            wife = Person.people.get(spouses["wife"])
+            person_list[ind].wife = wife
+            wife.husband = person_list[ind]
+        if "husband" in spouses and spouses["husband"] is not None:
```
Polyakiv-Andrey, line 22: is not None should be omitted, it will work without it

app/main.py
```diff
@@ -1,8 +1,26 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people.setdefault(name, self)
 
 
 def create_person_list(people: list) -> list:
```
Polyakiv-Andrey, line 10: add more detail type annotation for list

Polyakiv-Andrey (changes requested):

app/main.py
gesiod: I don't think so:
people = [{'age': 30, 'name': 'Ross', 'wife': 'Rachel'}, {'age': 29, 'name': 'Joey', 'wife': **None**}, ...

Without this additional condition, I don't pass the pytests.

app/main.py
gesiod: But it doesn't work)

app/main.py
gesiod: Done

Polyakiv-Andrey (approved):
Good job!

PR #821, opened 2023-10-25
'Solution'

aleksandra-prognimak:


app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🎈

PR #824, opened 2023-10-26
Solution

eLQeR:


app/main.py
```diff
@@ -1,8 +1,29 @@
...
+                for index2 in range(len(people)):
+                    if people[index1]["wife"] == people[index2]["name"]:
+                        person_list[index1].wife = person_list[index2]
+
+        elif "husband" in people[index1]:
+            if people[index1]["husband"] is not None:
+                for index2 in range(len(people)):
+                    if people[index1]["husband"] == people[index2]["name"]:
+                        person_list[index1].husband = person_list[index2]
+    return person_list
```
ArsenPidhoretskyi, line 29: So hard realization...... First of all why do you iterate through index? Also, why do you use loop into loop? 
You have Person.people dict use it and only one loop here

ArsenPidhoretskyi (changes requested):

app/main.py
eLQeR: I use the first loop to add attributes to each person and the second to find husband/wife for him, I don't know how to make it by one loop

eLQeR: I use index because it is the same in the dict and in the list or person and it is easy to iterate so -__-

eLQeR: Dont answer I got you )

YukihiroSM (approved):
Great job!

PR #828, opened 2023-10-27
Solution

Marinel444:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    married_dict = {}
+    result = []
+
+    for item in people:
+        person = Person(name=item["name"], age=item["age"])
+        married_dict[item["name"]] = person
+        result.append(person)
```
viktoria-rybenchuk, line 17: Consider using a list comprehension for a more concise approach

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    married_dict = {}
```
viktoria-rybenchuk, line 11: why do you need this dict?

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    result = []
+
+    for item in people:
+        person = Person(name=item["name"], age=item["age"])
+        married_dict[item["name"]] = person
+        result.append(person)
+
+    for item in people:
```
viktoria-rybenchuk, line 19: Consider using `person` instead of `item` for clarity

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+        person_obj = married_dict[item["name"]]
```
viktoria-rybenchuk, line 20: You can create a variable for 'person' and then use it, instead of adding it to the dictionary

viktoria-rybenchuk (changes requested):

app/main.py
Marinel444: to make a connection between married people

viktoria-rybenchuk (approved):
Well done!

PR #830, opened 2023-10-28
Solution

spryima:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
```
vasylhnatiuk, line 10:
```suggestion
def create_person_list(people: list) -> list[Person]:
```

app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    # write your code here
-    pass
+    persons = [Person(person_data["name"], person_data["age"])
+               for person_data in people]
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife \
```
vasylhnatiuk, line 16: try to avoid \ here

vasylhnatiuk (changes requested):

app/main.py
spryima: > try to avoid \ here

What should I use instead?

vasylhnatiuk (approved):

PR #833, opened 2023-10-31
solution_people

OleksandrHubar:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
-    pass
+    person_list = []
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person = Person(name, age)
+        spouse_name = person_data.get("wife") or person_data.get("husband")
+        if spouse_name:
+            spouse = Person.people.get(spouse_name)
+            if spouse:
```
Dimosphen1, line 20: Shorten the logic of these if conditions by combining them into one and avoiding additional nesting:
```suggestion
        if spouse_name and (spouse := Person.people.get(spouse_name)):
```

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+                person.spouse = spouse
+                if "wife" in person_data:
+                    person.wife = spouse
+                    spouse.husband = person
+        person_list.append(person)
+    return person_list
```
Dimosphen1, line 26: There is no need to have a `person_list` variable, you could extract values from the `people` class attribute, e.g.: 
```suggestion
    return list(Person.people.values())
```

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #835, opened 2023-11-01
Solution

KwintDimon:


app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    people_list = [Person(man["name"], man["age"]) for man in people]
```
viktoria-rybenchuk, line 13: use a more general name like `person` instead of specifying `man` or `woman`

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    for man in people:
```
viktoria-rybenchuk, line 14: fix here according to comment above

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #841, opened 2023-11-06
Solution

vpanchenkoSPS:


app/main.py
```diff
@@ -1,8 +1,19 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
fsocie7y, line 7: You should access class attribute through class name not self

app/main.py
```diff
@@ -1,8 +1,19 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    for index, person in enumerate(people):
+        if "wife" in person and person["wife"] is not None:
```
fsocie7y, line 13: Use just dict.get() method in the condition, instead of this long one

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+            person_list[index].wife = Person.people.get(person["wife"])
+
+        if "husband" in person and person["husband"] is not None:
```
fsocie7y, line 16: here to

fsocie7y (changes requested):

fsocie7y (approved):
GJ 👍

PR #843, opened 2023-11-10
Solution

halyna-baklanova:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(data: dict) -> list:
+    person_list = []
+
+    for person_data in data:
+        person = Person(person_data["name"], person_data["age"])
+        person_list.append(person)
```
pavlejviki, line 15: you can use list comprehension

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+
+    for person_data in data:
+        person = Person.people[person_data["name"]]
+        if "wife" in person_data and person_data["wife"] is not None:
```
pavlejviki, line 19: use get() method instead of two conditions

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+            person.wife = Person.people[person_data["wife"]]
+        elif "husband" in person_data and person_data["husband"] is not None:
```
pavlejviki, line 21: use get() method instead of two conditions

pavlejviki (changes requested):

tests/test_main.py
```diff
@@ -84,6 +84,7 @@ def test_create_person_list_has_wife_and_wife_have_husband(
 
 
 def test_create_person_list_has_no_wife(people_data, created_person_list):
+
```
Dimosphen1, line 87: Delete the changes made in this file, it should be absent in the PR.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(
+        data: list,
+        wife_key: str = "wife",
+        husband_key: str = "husband"
+) -> list:
```
Dimosphen1, line 14: Be more specific in the type annotation indicating a list of what entities you expect to obtain for the `data` and expect to return.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    person_list = [Person(person_data["name"],
+                   person_data["age"]) for person_data in data]
+
+    for person_data in data:
+        person = Person.people[person_data["name"]]
+
+        if person_data.get(wife_key) is not None:
+            wife_id = person_data.get(wife_key)
+            person.wife = Person.people.get(wife_id, None)
```
Dimosphen1, line 23: `None` is a default return value in `.get()`, there is no need to indicate this, so remove `None` from `.get()`, please.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(
+        data: list,
+        wife_key: str = "wife",
+        husband_key: str = "husband"
```
Dimosphen1, line 13: You do not need to dynamically define the wife and husband key name for this task, remove these arguments and hardcode values in the method itself.

app/main.py
Dimosphen1, line 23: There is an option two avoid extracting a value from the dict twice by using the walrus operator:
```suggestion
        if wife_name := person_data.get("wife"):
            person.wife = Person.people.get(wife_name)
```
Apply this change to a husband as well.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+) -> list:
+    person_list = [Person(person_data["name"],
+                   person_data["age"]) for person_data in data]
```
Dimosphen1, line 16: It is not a good practice to place two arguments that are used in an instance initialization on a new line in this way, place them on a same line, e.g.:
```suggestion
    person_list = [Person(person_data["name"], person_data["age"]) 
                   for person_data in data]
```

Dimosphen1 (changes requested):

pavlejviki (approved):
Good job!

PR #855, opened 2023-11-17
Solution

sofiamayatska:


app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    for person in people_list:
+        Person(person["name"], person["age"])
+
+    for person in people_list:
+        if "wife" in person and person["wife"] is not None:
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"] is not None:
+            Person.people[person["name"]].husband = Person.people[person["husband"]]
+
+    return list(Person.people.values())
```
ArsenPidhoretskyi, line 20: But if we two times run this function we should expect only people from people_list

app/main.py
```diff
@@ -1,8 +1,20 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
+    for person in people_list:
+        Person(person["name"], person["age"])
+
+    for person in people_list:
+        if "wife" in person and person["wife"] is not None:
```
ArsenPidhoretskyi, line 15: use .get instead of 2 conditions

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
+    Person.people.clear()
```
vsmutok, line 11: redundant here

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+    for person in people_list:
+        Person(person["name"], person["age"])
```
vsmutok, line 13: people = [Person (person["name"], person["age"]) for person in people_list]

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        person_instance = Person.people[person["name"]]
+        spouse_name = person.get("wife") or person.get("husband")
+
+        if spouse_name is not None:
+            if "wife" in person:
+                person_instance.wife = Person.people[spouse_name]
+            else:
+                person_instance.husband = Person.people[spouse_name]
+
+    return list(Person.people.values())
```
vsmutok, line 25:
```suggestion
    return people
```

vsmutok (changes requested):

PR #858, opened 2023-11-19
Solution

gertsovenko:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+        Person.people[name] = self
+
+
+def create_person_list(people_list: list) -> list:
+    new_people_list = []
+    for person in people_list:
+        new_people_list.append(Person(person["name"], person["age"]))
```
vsmutok, line 14: use list comprehension

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+                if woman.name == person["name"]:
+                    for man in new_people_list:
+                        if man.name == person["husband"]:
+                            woman.husband = man
+        if "wife" in person and person["wife"]:
+            for man in new_people_list:
+                if man.name == person["name"]:
+                    for woman in new_people_list:
+                        if woman.name == person["wife"]:
+                            man.wife = woman
```
vsmutok, line 27: try to optimize this algorithm

vsmutok (changes requested):

app/main.py
gertsovenko: Can you help? In what direction should I think?

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+    new_people_list = [Person(person["name"], person["age"])
+                       for person in people_list]
+
+    for person in people_list:
+        if "husband" in person and person["husband"]:
+            for woman in new_people_list:
+                for man in new_people_list:
+                    if (woman.name == person["name"]
+                            and man.name == person["husband"]):
+                        woman.husband = man
```
Dimosphen1, line 21: You could avoid the double loop iteration by changing the logic of the check. You could only check if there is a `"husband"` key and in case it is present, you could extract the person instance that is currently used in iteration, and extract the husband person instance by husband name from the `Person.people` class attribute. After that you could create the necessary `husband` and `wife` attributes in both instances. The same logic could be applied in case `"wife"` key is present.

Dimosphen1 (changes requested):
Change the logic of the `wife` and `husband` attributes assignment.

app/main.py
gertsovenko, line 1: Should I use variables wife and husband or it's better to write in one line?

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+            if person["name"] in Person.people:
+                husband = Person.people[person["husband"]]
+                wife = Person.people[person["name"]]
+                wife.husband = husband
+
+        elif "wife" in person and person["wife"]:
+            if person["name"] in Person.people:
+                wife = Person.people[person["wife"]]
+                husband = Person.people[person["name"]]
+                husband.wife = wife
```
Dimosphen1, line 26: You could simplify this condition in the following way:
```suggestion
    for person in people_list:
        person_instance = Person.people[person["name"]]
        if husband_name := person.get("husband"):
            person_instance.husband = Person.people[husband_name]
        elif wife_name := person.get("wife"):
             person_instance.wife = Person.people[wife_name]
```
Modify this approach in case there are errors during the program execution but keep the overall structure.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
-    pass
+        Person.people[name] = self
+
+
+def create_person_list(people_list: list) -> list:
+    new_people_list = [Person(person["name"], person["age"])
+                       for person in people_list]
+    for person in people_list:
+        person_instance = Person.people[person["name"]]
+        if "husband" in person and person["husband"]:
```
viktoria-rybenchuk, line 16: use  the `get` method for better readability and safety

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+            person_instance.husband = Person.people[person["husband"]]
+        elif "wife" in person and person["wife"]:
```
viktoria-rybenchuk, line 18: same here

app/main.py
```diff
@@ -1,8 +1,20 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+        Person.people[name] = self
+
+
+def create_person_list(people_list: list) -> list:
+    new_people_list = [Person(person["name"], person["age"])
+                       for person in people_list]
```
viktoria-rybenchuk, line 13:
```suggestion
    new_people_list = [
    Person(person["name"], person["age"])
     for person in people_list
     ]
```

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #862, opened 2023-11-20
Class Person implemented

Black3942:


fsocie7y (changes requested):
how about person husband? i think u need one more condition to check)

PR #864, opened 2023-11-21
Solution

Vlad1slavo:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
     pass
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
+    for person in people:
+        Person(person["name"], person["age"])
+    for person in people:
+        if person.get("wife") is not None:
+            Person.people[person["name"]].wife\
```
vasylhnatiuk, line 16: try to avoid \

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+                = Person.people[person["wife"]]
+        elif person.get("husband") is not None:
+            Person.people[person["name"]].husband\
```
vasylhnatiuk, line 19: the same here

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
     pass
 
 
 def create_person_list(people: list) -> list:
```
vasylhnatiuk, line 11:
```suggestion
def create_person_list(people: list) -> list[Person]:
```

vasylhnatiuk (changes requested):

app/main.py
Vlad1slavo: > the same here

Corrected

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+    for person in people:
+        current_person = Person.people[person["name"]]
+        if person.get("wife") is not None:
+            wife = Person.people[person["wife"]]
+            current_person.wife = wife
+        elif person.get("husband") is not None:
+            husband = Person.people[person["husband"]]
+            current_person.husband = husband
+
+    return list(Person.people.values())
```
ArsenPidhoretskyi, line 23: This is not true. If we two times invoke this function we do not need previous results

ArsenPidhoretskyi: Return created objects into function

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,8 +1,17 @@
...
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list) -> list[Person]:
+    person_list = [Person(pers["name"], pers["age"]) for pers in people]
+    for pers, person in zip(people, person_list):
+        if "wife" in pers and pers["wife"]:
```
ArsenPidhoretskyi, line 13: use .get instead of two conditions

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #901, opened 2023-12-25
Solution

kabust:


msymonovych (approved):

app/main.py
```diff
@@ -1,8 +1,17 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    for index, person in enumerate(people):
+        if "wife" in person and person["wife"] is not None:
+            person_list[index].wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"] is not None:
```
NEkropo1, line 15: You can simplify this by removing the is not None check. If person["wife"] or person["husband"] is None, or if these keys are not present in the dictionary, the conditional statement will evaluate to False. This simplification works under the assumption that other falsy values (like empty strings, zeros, or empty lists) are not valid or expected values for these keys.

NEkropo1 (changes requested):

NEkropo1 (approved):
LGTM!

PR #916, opened 2024-01-03
Solution

gaupt:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        self.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = list()
+    for person_data in people:
+        person = Person(person_data["name"], person_data["age"])
+    for person in people:
```
YuriiSmolii, line 14: This loop is identical with the first one so that can be simplified in order to reduce the complexity of the algorithm.

YuriiSmolii (changes requested):
Algorithm simplification change requested

app/main.py
gaupt: > This loop is identical with the first one so that can be simplified in order to reduce the complexity of the algorithm.

not work

gaupt:
Fix

NazarSheremet (approved):
Good job!

PR #917, opened 2024-01-03
Solution

oksankakot:


kabust (approved):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+    for i in range(len(person_list)):
+        person = person_list[i]
+        person_dict = people[i]
+        spouse = person_dict.get("wife") or person_dict.get("husband")
+        if spouse is not None:
+            spouse_person = Person.people[spouse]
+            if "wife" in person_dict:
+                person.wife = spouse_person
+            else:
+                person.husband = spouse_person
```
NEkropo1, line 24: best case would be to create function from that and make good returns, but it's not necessary here)

NEkropo1 (approved):
Good Job!

PR #918, opened 2024-01-03
Solution

serji13:


EdAlekseiev (commented):
Good job

serji13:
> Good job

If it's good job, can you approve the changes?

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    list_of_people = []
+    for d_person in people:
+        c_person = Person(d_person.get("name"), d_person["age"])
+        if "husband" in d_person:
+            if d_person.get("husband") is not None:
+                c_person.husband = d_person.get("husband")
+        if "wife" in d_person:
+            if d_person.get("wife") is not None:
+                c_person.wife = d_person.get("wife")
+        list_of_people.append(c_person)
```
YukihiroSM, line 20: It's better to just create peoples here without getting husband/wife and after that you will have Person.people - which will contain all peoples. After that in the second loop you can iterate over peoples and you will set at the ready Person objeccts as husband/wife without trying to understand if this is a string or something else.

YukihiroSM (changes requested):

app/main.py
serji13: I did not quite understand how I would be able to fit in two cycles in this way. But I improved my previous code - hope you like it.

EdAlekseiev (approved):
Good job

PR #924, opened 2024-01-06
Solution

MykhailoIvchenko:
In the function add_person_to_the_list for the "person" argument annotation the "object" was used because tests fail if I use Person with import annotation from __future__

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,32 @@
 class Person:
-    # write your code here
-    pass
+    people: dict = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.add_person_to_the_list(self)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,32 @@
...
+    people: dict = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.add_person_to_the_list(self)
+
+    @classmethod
+    def add_person_to_the_list(cls, person: object) -> None:
+        cls.people[person.name] = person
```
YukihiroSM, line 12: BTW, it can be simlified to Person.people[name] = self, but your solution is good as well.

YukihiroSM (approved):

PR #954, opened 2024-01-14
Sol

Dobryk12:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
+
+    for i in people_list:
+        name = i["name"]
+        age = i["age"]
+        person = Person(name, age)
+        persons.append(person)
+
+    for person in people_list:
+        if "wife" in person:
+            if person["wife"] is not None:
```
Anna728560, line 21: you can use dict get() method to simplify this

Anna728560 (changes requested):

app/main.py
NEkropo1: Actually if there's no key adressing the key of dict will throw an error, while method `get()` will just get None. And if you want to make it python way you should write:
`if person.get("wife"):`
which will make all detections of empty value and None.

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        self.__class__.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
+    persons = []
+
+    for i in people_list:
```
NEkropo1, line 13: what is i?

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        age = i["age"]
+        person = Person(name, age)
+        persons.append(person)
+
+    for person in people_list:
+        if person.get("wife") is not None:
+            way_to_person = Person.people[person["name"]]
+            way_to_person.wife = Person.people[person["wife"]]
+        if "husband" in person:
+            if person["husband"] is not None:
```
NEkropo1, line 24: simply `if person.get("wife"):`

NEkropo1 (changes requested):

NEkropo1 (changes requested):
You should reflect changes on all your code, not only on places, where someone comments )

PR #956, opened 2024-01-15
'Solution'

sofiazholud:


app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if "wife" in person and person["wife"]:
```
EdAlekseiev, line 14:
```suggestion
        if person.get("wife"):
```

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"]:
```
EdAlekseiev, line 16:
```suggestion
        if person.get("husband"):
```

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+            Person.people[person["name"]].husband = (
+                Person.people)[person["husband"]]
```
EdAlekseiev, line 18:
```suggestion
                Person.people[person["husband"]])
```

EdAlekseiev (changes requested):

app/main.py
sofiazholud: solve it

app/main.py
sofiazholud: solved it

app/main.py
sofiazholud: solved it

EdAlekseiev (approved):
Good job

PR #968, opened 2024-01-20
Solution

yaaszp:


app/main.py
```diff
@@ -1,8 +1,34 @@
...
+        if None in dic.values():
+            res_list.append(Person(dic["name"], dic["age"]))
+        elif "wife" in dic.keys():
+            person = Person(dic["name"], dic["age"])
+            person.wife = dic["wife"]
+            res_list.append(person)
+        elif "husband" in dic.keys():
+            person = Person(dic["name"], dic["age"])
+            person.husband = dic["husband"]
+            res_list.append(person)
```
NazarSheremet, line 23: There is no need for a two-time search for wife and husband. By the first iteration of the people list need to create an object of class without a partner and by the second iteration add them

NazarSheremet (changes requested):
Need to optimize the solution

app/main.py
yaaszp: optimize done

NazarSheremet (changes requested):
Need some changes

app/main.py
```diff
@@ -1,8 +1,24 @@
...
-    # write your code here
-    pass
+    res_list = []
+
+    for dic in people:
+        res_list.append(Person(dic["name"], dic["age"]))
+
+    for dic in people:
+        if None in dic.values():
+            continue
```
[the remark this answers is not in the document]
yaaszp: Because, without this "if" code doesn't work

YuriiSmolii (approved):
LGTM

PR #978, opened 2024-01-25
Solution

vsupruniuk:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
+    people: dict = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+        self.__class__.people[name] = self
```
LiudmylaKulinchenko, line 8: It's common practice to refer to class attributes directly by the class name rather than through `self.__class__`. This makes the code more readable and clear about which attribute is being accessed

When dealing with inheritance, using `self.__class__` to reference a class attribute can be beneficial because it ensures that the attribute of the current class is used, not the one from the parent class

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    for person in people:
+        Person(person["name"], person["age"])
```

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+
+    for person in people:
+        if "wife" in person and person["wife"] in Person.people:
```

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+            wife = Person.people[person["wife"]]
+            husband = Person.people[person["name"]]
+
+            wife.husband = husband
+            husband.wife = wife
+
+        if "husband" in person and person["husband"] in Person.people:
```
LiudmylaKulinchenko, line 23: same here about .get()

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+            wife = Person.people[person["name"]]
+            husband = Person.people[person["husband"]]
+
+            wife.husband = husband
+            husband.wife = wife
+
+    return [person for person in Person.people.values()]
```
LiudmylaKulinchenko, line 30: You can save in the variable a list that you create at the start of function and return it here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🌟

PR #979, opened 2024-01-25
Solution

EugeneProtsenko:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    persons = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = \
+                Person.people[person["wife"]]
```
EdAlekseiev, line 20: Use `()` instead of `\` for separating line

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        elif person.get("husband"):
+            Person.people[person["name"]].husband = \
+                Person.people[person["husband"]]
```
EdAlekseiev, line 23: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #982, opened 2024-01-26
solution with some problems

prev1ew:
Idk how to solve 3 "not passed" tests

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -4,65 +4,38 @@ from __future__ import annotations
 class Person:
     people = {}  # people collection, key=name, value = ref to a person
 
-    def __init__(self, name: str, age: int, wife: str = None, husband: str = None) -> None:
-        # fill data
+    def __init__(self, name: str, age: int) -> None:
+
         self.name = name
         self.age = age
 
-        if wife:
-            self.wife = self.find_partner(wife)
-        if husband:
-            self.husband = self.find_partner(husband)
-
-        self.people[self.name] = self
-
-    def find_partner(self, person_to_find: str | None) -> Person | str | None:
-        # if the person exists
-        if person_to_find and person_to_find in self.people.keys():
-            partner = self.people[person_to_find]
-            # write this person as a partner`s partner
-            if hasattr(partner, "wife") and partner.wife == self.name:
-                partner.wife = self
-            elif hasattr(partner, "husband") and partner.husband == self.name:
-                partner.husband = self
-            return self.people[person_to_find]
-        else:  # return str if partner non exists
-            return person_to_find
+        Person.people[self.name] = self
 
 
-def create_person_list(people: list) -> list:
-    return [Person(
-            name=human_being["name"],
-            age=human_being["age"],
-            wife=human_being["wife"] if "wife" in human_being.keys() else None,
-            husband=human_being["husband"] if "husband" in human_being.keys() else None)
-            for human_being in people]
+def find_partner(obj: Person,
+                 person_to_find: str | None
+                 ) -> Person | str | None:
+    # if the person exists
+    if person_to_find and person_to_find in obj.people.keys():
+        partner = obj.people[person_to_find]
+        # write this person as a partner`s partner
+        if hasattr(partner, "wife") and partner.wife == obj.name:
+            partner.wife = obj
+        elif hasattr(partner, "husband") and partner.husband == obj.name:
+            partner.husband = obj
+        return obj.people[person_to_find]
+    else:  # return str if partner non exists
+        return person_to_find
 
 
-test = [
-        {"name": "Ross", "age": 30, "wife": "Rachel"},
-        {"name": "Joey", "age": 29, "wife": None},
-        {"name": "Phoebe", "age": 31, "husband": None},
-        {"name": "Chandler", "age": 30, "wife": "Monica"},
-        {"name": "Monica", "age": 32, "husband": "Chandler"},
-        {"name": "Rachel", "age": 28, "husband": "Ross"},
-    ]
+def create_person_list(people: [dict]) -> [Person]:
 
-# lst = create_person_list(test)
-# for index, persn in enumerate(lst, 0):
-#     print(f"{index}. {persn.name}")
-#
-# # 0. Ross <__main__.Person object at 0x000001BE361F85C0> None
-# # 1. Joey None None
-# # 2. Phoebe None None
-# # 3. Chandler <__main__.Person object at 0x000001BE361F8560> None
-# # 4. Monica None <__main__.Person object at 0x000001BE361F84D0>
-# # 5. Rachel None <__main__.Person object at 0x000001BE361F8440>
-#
-# print(lst[0].wife.name)  # Rachel
-# print(lst[5].husband.name)  # Ross
-#
-# print(lst[0].wife is lst[5] and lst[5].husband is lst[0])  # true
-#
-# print("*"*20)
+    for human_data in people:
+        curr_person = Person(human_data["name"], human_data["age"])
+        if "wife" in human_data.keys():
+            curr_person.wife = find_partner(curr_person, human_data["wife"])
+        elif "husband" in human_data.keys():
+            curr_person.husband = find_partner(curr_person,
+                                               human_data["husband"])
 
+    return list(Person.people.values())
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,68 @@
+from __future__ import annotations
+
+
 class Person:
-    # write your code here
-    pass
+    people = {}  # people collection, key=name, value = ref to a person
+
+    def __init__(self, name: str, age: int, wife: str = None, husband: str = None) -> None:
+        # fill data
+        self.name = name
+        self.age = age
+
+        if wife:
+            self.wife = self.find_partner(wife)
+        if husband:
+            self.husband = self.find_partner(husband)
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,36 @@
...
 
+    def __init__(self, name: str, age: int) -> None:
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+        self.name = name
+        self.age = age
+
+        self.people[self.name] = self
```
pavlejviki, line 9: use Person to access class attributes instead of self

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        return person_obj.people[person_to_find]
+    else:
+        return person_to_find
+
+
+def create_person_list(people: [dict]) -> [Person]:
+
+    for human_data in people:
+        curr_person = Person(human_data["name"], human_data["age"])
+        if "wife" in human_data.keys() and human_data["wife"]:
```
pavlejviki, line 30: you don't need to call keys() , `in` does it under the hood

and better use get() method instead of two conditions

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+                 ) -> Person | str | None:
+    if person_to_find and person_to_find in person_obj.people.keys():
+        partner = person_obj.people[person_to_find]
+        if hasattr(partner, "wife"):
+            partner.wife = person_obj
+        elif hasattr(partner, "husband"):
+            partner.husband = person_obj
+        return person_obj.people[person_to_find]
+    else:
+        return person_to_find
```
pavlejviki, line 23: what is the point to return a string ? you are supposed to assign a Person instance to the wife/husband attributes

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+def find_partner(person_obj: Person,
+                 person_to_find: str | None
+                 ) -> Person | str | None:
+    if person_to_find and person_to_find in person_obj.people.keys():
+        partner = person_obj.people[person_to_find]
+        if hasattr(partner, "wife"):
+            partner.wife = person_obj
+        elif hasattr(partner, "husband"):
+            partner.husband = person_obj
+        return person_obj.people[person_to_find]
```
pavlejviki, line 21: When accessing a class attribute , it is generally considered best practice to use the class name instead of an instance (self) to access that attribute.

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        self.name = name
+        self.age = age
+
+        self.people[self.name] = self
+
+
+def find_partner(person_obj: Person,
+                 person_to_find: str | None
+                 ) -> Person | str | None:
+    if person_to_find and person_to_find in person_obj.people.keys():
```
pavlejviki, line 15: keys() is redundant

pavlejviki (changes requested):

app/main.py
prev1ew: thank you for answering question "why", i didn't know it

app/main.py
```diff
@@ -1,8 +1,34 @@
...
+    if person_to_find in Person.people:
+        partner = person_obj.people[person_to_find]
+        if hasattr(partner, "wife"):
+            partner.wife = person_obj
+        elif hasattr(partner, "husband"):
+            partner.husband = person_obj
+        return Person.people[person_to_find]
+
+
+def create_person_list(people: [dict]) -> [Person]:
```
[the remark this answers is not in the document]
prev1ew: list with dicts, list with Persons, like [{}] and list(Person)

app/main.py
```diff
@@ -1,8 +1,34 @@
...
+def find_partner(person_obj: Person,
+                 person_to_find: str | None
+                 ) -> Person | str | None:
+    if person_to_find in Person.people:
+        partner = person_obj.people[person_to_find]
+        if hasattr(partner, "wife"):
+            partner.wife = person_obj
+        elif hasattr(partner, "husband"):
+            partner.husband = person_obj
+        return Person.people[person_to_find]
```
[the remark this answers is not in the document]
prev1ew: why?

app/main.py
```diff
@@ -1,8 +1,34 @@
...
+
+def create_person_list(people: [dict]) -> [Person]:
+
+    for human_data in people:
+        curr_person = Person(human_data["name"], human_data["age"])
+        if human_data.get("wife"):
+            curr_person.wife = find_partner(curr_person, human_data["wife"])
+        elif human_data.get("husband"):
+            curr_person.husband = find_partner(curr_person,
+                                               human_data["husband"])
```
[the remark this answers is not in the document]
prev1ew: Why should I use the suggested method? This problem can be solved with different methods. I do not know "best practice" and would love to hear your explanation... In other words - please, answer the question "why?"

app/main.py
```diff
@@ -1,8 +1,34 @@
...
+
+    return list(Person.people.values())
```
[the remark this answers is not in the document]
prev1ew: The function should return a list of objects. It returns objects. Please, can you tell me more about the problem, right now - I can't see a reason to change the code


Ofc it will return you different lists if you call it twice, is It a problem? I can't see "this" specified as a "must" in the description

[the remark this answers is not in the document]
prev1ew: I can change it, but I want to know the reason

app/main.py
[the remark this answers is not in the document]
prev1ew: is it better to use "list" instead of "[]" ?

app/main.py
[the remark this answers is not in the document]
prev1ew: I don't know what you want from me, please write in detail what you are expecting

app/main.py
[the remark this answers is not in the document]
prev1ew: It's bad practice to loop items twice when I can do this with one loop, I disagree with your suggestion.
But I see your point, I'll create a new function to set a partner

app/main.py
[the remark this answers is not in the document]
prev1ew: no, I can't, due to tests I can't add a new variable "people_list"

app/main.py
```diff
@@ -1,8 +1,34 @@
...
 
+    def __init__(self, name: str, age: int) -> None:
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+        self.name = name
+        self.age = age
+
+        Person.people[self.name] = self
```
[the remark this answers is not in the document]
prev1ew: Why? 
Other reviewers told me "It's bad practice to access class variables with "self".
I can't see when I should access class variables with "self" and when with "<ClassName>". 
Please, explain

I will fix it when I find out when to use "self.<ClassAttr>" and "<ClassName>.<ClassAtr>

prev1ew:
sorry, now I see that you want me to create a new function

prev1ew:
I re-read your reply, now I see that it should return not all Persons but only created, I'll fix it soon

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+
+
+def find_partner(person_obj: Person,
+                 person_to_find: str | None
+                 ) -> Person | str | None:
```
Dimosphen1, line 14: Place all arguments on a new line, do not place arguments on the same line with the opening parenthesis, it should not have any arguments on the same line as the closing parenthesis.

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+    if person_to_find in Person.people:
+        partner = person_obj.people[person_to_find]
```
Dimosphen1, line 16: Call `people` class attribute from the `Person` class as it is done on line 21, not from the class instance.

app/main.py
Dimosphen1, line 14: Revise your return type annotation, there is no case when `str` will be returned.

app/main.py
```diff
@@ -1,8 +1,38 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+        self.name = name
+        self.age = age
+
+        Person.people[self.name] = self
+
+
+def find_partner(person_obj: Person,
```
Dimosphen1, line 12: Your function name is ambiguous since the function itself does not only find a partner but also sets the `wife` or `husband` attribute to it. Either rename the function or modify its logic.

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+                 person_to_find: str | None
+                 ) -> Person | str | None:
+    if person_to_find in Person.people:
+        partner = person_obj.people[person_to_find]
+        if hasattr(partner, "wife"):
+            partner.wife = person_obj
+        else:  # hasattr(partner, "husband"):
```
Dimosphen1, line 19: There is no need for this comment, you could make the following condition
```suggestion
        elif hasattr(partner, "husband"):
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,8 +1,44 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+        self.name = name
+        self.age = age
+
+        Person.people[self.name] = self
+
+
+def find_partner_and_set_his_partner(
```
Dimosphen1, line 12: You could use the `get_and_update_partner` name here.

Dimosphen1 (approved):
LGTM, all requirements are met and there are no code style issues. Please, request a new review from the mentor after all changes are pushed and the code is in its final version without pushing new changes after the review request.

PR #988, opened 2024-01-29
'Solution'

ValeriyIvanovich:


app/main.py
```diff
@@ -1,8 +1,20 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
Dimosphen1, line 7: `people` is a class attribute, call it from `Person`, not `self`.

app/main.py
```diff
@@ -1,8 +1,20 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_person = [Person(person["name"],
+                             person["age"]) for person in people]
```
Dimosphen1, line 12: It would be better to format the list comprehension in a way to leave the `Person` instance initialization on one line, e.g.
```suggestion
    list_of_person = [Person(person["name"], person["age"]) 
                      for person in people]
```

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    for i, person in enumerate(people):
```
Dimosphen1, line 13: Avoid using single-letter variable names in your code, use `index` instead.

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        if person.get("husband"):
+            husband_name = person["husband"]
```
Dimosphen1, line 15: You could use a walrus operator to not perform the operation of retrieving name twice during the check and variable assignment:
```suggestion
        if husband_name := person.get("husband"):
```

Dimosphen1 (changes requested):

ValeriyIvanovich:
Thank. All suggested changes have been implemented

pavlejviki (approved):
Looks good

PR #990, opened 2024-01-30
Solution

TarasFirst:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+
+    for one in people:
+        Person(one["name"], one["age"])
+
+    for one in people:
+        if "wife" in one and one["wife"] is not None:
+            Person.people[one["name"]].wife = Person.people[one["wife"]]
+
+        elif "husband" in one and one["husband"] is not None:
+            Person.people[one["name"]].husband = Person.people[one["husband"]]
```
Dimosphen1, line 21: Use a more descriptive variable naming instead of `one`, e.g. `person_data`.

app/main.py
```diff
@@ -1,8 +1,23 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    for one in people:
+        Person(one["name"], one["age"])
+
+    for one in people:
+        if "wife" in one and one["wife"] is not None:
+            Person.people[one["name"]].wife = Person.people[one["wife"]]
```
Dimosphen1, line 18: You could simplify the if condition by using `get`:
```suggestion
        if one.get("wife"):
            Person.people[one["name"]].wife = Person.people[one["wife"]]
```
Besides that, you could use walrus operator to retrieve, evaluate and save the name:
```suggestion
        if wife_name := one.get("wife"):
            Person.people[one["name"]].wife = Person.people[wife_name]
```

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    for one in people:
+        Person(one["name"], one["age"])
```
Dimosphen1, line 14: It would be better to make a list comprehension here.

app/main.py
Dimosphen1, line 21: You could retrieve the `Person` instance at the beginning of an iteration instead of calling it twice in if condition:
```suggestion
    for one in people:
        person_instance =  Person.people[one["name"]]
        if "wife" in one and one["wife"] is not None:
            person_instance.wife = Person.people[one["wife"]]

        elif "husband" in one and one["husband"] is not None:
            person_instance.husband = Person.people[one["husband"]]
```

Dimosphen1 (changes requested):

app/main.py
TarasFirst: this way i tried to reduce a lenght of the string of code for flake8 ))
how i can do it another way ?

PR #999, opened 2024-02-02
added solution

MAFIOZNIK1999:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    pupil_array = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = \
```
LiudmylaKulinchenko, line 15: Better not to use ` \` as a linebreaker, create temporary variable instead or use parentheses
While both `\` and parentheses can be used for line continuation, parentheses are generally recommended for their clarity, readability, and predictability

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 👻

PR #1006, opened 2024-02-05
Solution

IhorDmytriv:


app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = []
+    for person in people:
+        people_list.append(Person(name=person["name"], age=person["age"]))
```
NikOneZ1, line 13: You can use list comprehension here

NikOneZ1 (changes requested):
It would be better to use .get() method instead of the []

PR #1013, opened 2024-02-07
add task solution

EugeneSherepa:


app/main.py
```diff
@@ -1,8 +1,33 @@
...
-    # write your code here
-    pass
+    people_list = []
+
+    for person in people:
+        name = person["name"]
+        age = person["age"]
+        new_person = Person(name, age)
+        people_list.append(new_person)
+
```
LiudmylaKulinchenko, line 18: U can create list in one line using list comprehension

app/main.py
```diff
@@ -1,8 +1,33 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+
+    for person in people:
+        name = person["name"]
+        age = person["age"]
+        new_person = Person(name, age)
+        people_list.append(new_person)
+
+        if "wife" in person:
+            wife_name = person["wife"]
+            if wife_name in Person.people:
```
LiudmylaKulinchenko, line 21: It's better to use `and` instead of nested `if` blocks
Also read about dict.get() method, it will be even better in this case

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,24 @@
...
-    pass
+    people_list = [
+        Person(
+            person["name"],
+            person["age"]
+        )
+        for person in people
+    ]
+    for person in people:
+        if wife_name := person.get("wife"):
```
LiudmylaKulinchenko, line 19: Interesting choice 👏

LiudmylaKulinchenko (approved):
Good job ☀

PR #1021, opened 2024-02-14
create class Person and function that adding objects for people

include031:


app/main.py
```diff
@@ -1,8 +1,25 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
Dimosphen1, line 7: `people` is a class attribute, call it from `Person` class, not `self`

app/main.py
```diff
@@ -1,8 +1,25 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        person_temp = Person(person["name"], person["age"])
+        person_list.append(person_temp)
```
Dimosphen1, line 14: Reformat this code to use a list comprehension, please

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+
+    for person in people:
+        if "wife" in person:
+            if person["wife"] is not None:
```
Dimosphen1, line 18: You could avoid nesting and use `get` to perform the check
```suggestion
        if person.get("wife"):
```
Besides that, you could save the retrieved value to a variable using walrus operator as well
```suggestion
        if wife_name := person.get("wife"):
```

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+                Person.people[person["name"]].wife = \
```
Dimosphen1, line 19: You could avoid line breaking in case you will retrieve a person instance at the beginning of the loop instead of referring to `Person.people[person["name"]]` in the if condition

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,8 +1,26 @@
...
-    pass
+    person_list = [
+        Person(person["name"], person["age"])
+        for person in people
+    ]
+
+    Person.people = {
+        person.name: person
+        for person in person_list
+    }
```
pavlejviki, line 18: you can implement the logic of adding the newly created instance to the Person.people dict in the __init__ method

pavlejviki (changes requested):

app/main.py
include031: def __init__(self, name: str, age: int) -> None:
        self.name = name
        self.age = age
        self.people[name] = self

it's correct?

app/main.py
include031: def init(self, name: str, age: int) -> None:
self.name = name
self.age = age
self.people[name] = self

Polyakiv-Andrey (approved):
Good job!

PR #1051, opened 2024-02-25
Class Person implemented

antonver:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    persons_list = [Person(man["name"], man["age"]) for man in humans]
+    for num in range(len(humans)):
+        if humans[num].get("husband"):
+            for man in persons_list:
+                if man.name == humans[num].get("husband"):
+                    persons_list[num].husband = man
+        elif humans[num].get("wife"):
+            for woman in persons_list:
+                if woman.name == humans[num].get("wife"):
+                    persons_list[num].wife = woman
```
Dimosphen1, line 20: Rework your logic to avoid using three loops (use just one and retrieve instances by key instead of iteration) and do not use shortened variable names such as `num`

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.name = name
+        self.age = age
+        self.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(humans: list) -> list:
+    persons_list = [Person(man["name"], man["age"]) for man in humans]
```
Dimosphen1, line 11: `man` variable name is not accurate replace is with `person`

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(humans: list) -> list:
```
Dimosphen1, line 10: There is no need to redefine the argument, return back previous function signature

app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
Dimosphen1, line 7: `people` is a class attribute, call it from `Person` instead of `self`

Dimosphen1 (changes requested):

PR #1052, opened 2024-02-26
Solution

IjdmLoverI:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
+    people = {}
 
+    def __init__(
+            self,
+            name: str,
+            age: int
+    ) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
```
Dimosphen1, line 11: Call `people` from `Person` class directly without using dunder methods

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    person_instances = {}
+    for person_info in people:
+        person_name = person_info["name"]
+        person = Person(person_name, person_info["age"])
+        person_instances[person_name] = person
+
+    for person_info in people:
+        person_name = person_info["name"]
```
Dimosphen1, line 22: Retrieve not a person name, but a person instance here
```suggestion
        person_instance = Person.people[person_info["name"]]
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        Person.people[self.name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    Person.people = {person_info["name"]: Person(person_info["name"],
+                                                 person_info["age"])
+                     for person_info in people}
```
Dimosphen1, line 17: There is no need to perform dict comprehension and assign values to `Person.people`. Make a list comprehension with the creation of instances which will automatically populate `Person.people` since this operation is done in the `__init__`
```suggestion
    [Person(person_info["name"], person_info["age"])
     for person_info in people]
```

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+    for person_info in people:
+        person_instance = Person.people[person_info["name"]]
+        if "wife" in person_info and \
+                (wife_name := person_info["wife"]):
```
Dimosphen1, line 21: Use `get()` to avoid checking for the presence of the key in the `person_info`, the loop will not enter if condition in case value is absent by this key
```suggestion
        if wife_name := person_info.get("wife"):
```

Dimosphen1 (changes requested):

app/main.py
IjdmLoverI: this doesnt work

PR #1066, opened 2024-03-09
Solution

Bujhm99:


Boyok1337 (changes requested):
remove comments

LiudmylaKulinchenko (changes requested):
All checks should pass

Bujhm99:
I did some manipulations with branches, so all changes in othe branch Dev2. I tried to push from pychar new solution by branch name Dev1, but looks like i created new PR (i don`t remember). What to do now? Del all, make new?

PR #1074, opened 2024-03-14
Second_try

Sydorovich:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
-    # write your code here
-    pass
+    result = [Person(person["name"], person["age"]) for person in people]
+    i = 0
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
+            result[i].wife = Person.people[person["wife"]]
+        if "husband" in person and person["husband"] is not None:
+            result[i].husband = Person.people[person["husband"]]
+        i += 1
```
NazarSheremet, line 18: `i` var is useless you can do a better way, also no need to write `is not None`
```suggestion
    for person in people:
        if "wife" in person and person["wife"]:
            Person.people[person.get("name")].wife = Person.people[person.get("wife")]
        if "husband" in person and person["husband"] is not None:
            Person.people[person.get("name")].husband = Person.people[person["husband"]]
```

NazarSheremet (changes requested):
Need changes

PR #1077, opened 2024-03-15
Solution for PY-PERSON-CLASS

NinqueRaven13:
There is one problem that I really did not solve. I would like to have a beautiful type annotation for __init__
image
To make it beautiful. I tried to use annotations from __future__. It makes code beautiful, but after that I got trouble with test
image
Please give piece of advice how make meaningful type annotation and pass that test

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,36 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_created_people = []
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,27 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_created_people = []
+    for person in people:
+        inst_person = Person(person["name"], person["age"])
+        if "wife" in person.keys():
+            key = "wife"
```
vasylhnatiuk, line 15:
```suggestion
            partner = "wife"
```

vasylhnatiuk: it is a better name

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1079, opened 2024-03-15
Solution

DmytroHlazyrin:


Bujhm99 (approved):
LGTM

app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
YuriiKindrat, line 7: Why don't u just use a class name here ?

app/main.py
```diff
@@ -1,8 +1,21 @@
...
-    pass
+    persons_data = [Person(person["name"], person["age"]) for person in people]
+    for person in people:
+
+        if person.get("wife"):
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+
+        if person.get("husband"):
+            Person.people[person["name"]].husband\
+                = Person.people[person["husband"]]
```
YuriiKindrat, line 19: could u break the line here, without using linebreakers ?

YuriiKindrat (changes requested):

Oleksl888 (approved):

PR #1086, opened 2024-03-17
Solution

panicua:


Bujhm99 (approved):
LGTM

panicua:
I see that I added few blank lines in the test_main.py file, but is it even worth to create a new commit to remove it?

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        person_by_name = Person(person["name"], person["age"])
+        people_by_name.append(person_by_name)
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife \
+                = Person.people[person["wife"]]
+        if person.get("husband"):
+            Person.people[person["name"]].husband \
+                = Person.people[person["husband"]]
```
YuriiKindrat, line 23: it is better to avoid using linebreakers 
could u break the line here without them ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        person_by_name = Person(person["name"], person["age"])
+        people_by_name.append(person_by_name)
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = (
+                Person.people)[person["wife"]]
+        if person.get("husband"):
+            Person.people[person["name"]].husband = (
+                Person.people)[person["husband"]]
```
YuriiKindrat, line 23: bad formmating here (
please, use a code style as mentioned in this file:
https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md

YuriiKindrat (changes requested):

app/main.py
panicua: There are not examples with parentheses on the URL, if you are talking about them, I'll change now.

YuriiKindrat: `Person.people[`
`   person["name"]`
`].husband = Person.people[person["husband"]]`
f.e. u could break the line like this
or like this:
`Person.people[ person["name"]].husband = Person.people[`
`    person["husband"]`
`]`

YuriiKindrat (approved):
LGTM, but read the comments)

PR #1092, opened 2024-03-18
Solution

Gilgumish:


mikhailyemets (approved):
gj

app/main.py
```diff
@@ -1,8 +1,20 @@
...
-    # write your code here
-    pass
+    person_list = [Person(info["name"], info["age"]) for info in people]
+
+    for info in people:
+        result = Person.people.get(info["name"])
+        if info.get("wife") and info["wife"] is not None:
+            result.wife = Person.people.get(info["wife"])
+        elif info.get("husband") and info["husband"] is not None:
+            result.husband = Person.people.get(info["husband"])
```
YuriiKindrat, line 18: it is better to use call var 'preson' instead of info, for better readability

YuriiKindrat (approved):

PR #1094, opened 2024-03-18
Solution

SashaKisliy:


oleksandra-shershen (approved):
LGTM

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list) -> list[Person]:
+    peoples = [Person(man["name"], man["age"]) for man in people]
```
YuriiKindrat, line 11: maybe it is better to call var person ?)

YuriiKindrat (approved):

app/main.py
SashaKisliy: Agreed)

PR #1096, opened 2024-03-18
Solution

SerhiiMar:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
-    pass
+    people = dict()
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
```
SashaKisliy, line 10: u must return list[Person]

SashaKisliy (approved):
Nice job!

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+def create_person_list(people: list) -> list[Person]:
+    def get_instance(name: str) -> Person:
+        for person in person_instances:
+            if person.name == name:
+                return person
+
+    person_instances = [
+        Person(person["name"], person["age"])
+        for person in people
+    ]
```
Oleksl888, line 19: This can fit on one line

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+
+    for index, person in enumerate(people):
+        if person.get("wife", None) is not None:
```
Oleksl888, line 22: There is no need to check 'is not None' here - person.get("wife") will return truthy or falsy value anyway

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+            person_instances[index].wife = get_instance(person["wife"])
```
Oleksl888, line 23: You do not need nested cycles here - it slows your program down. Try using one iteration and accessing objects through a class. This way you won't need use enumerate and get_instance functions as well

Oleksl888 (changes requested):

app/main.py
SerhiiMar: flake8 said that line was too long so i fitted it on 2 lines

app/main.py
SerhiiMar: Done

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+def create_person_list(people: list) -> list[Person]:
+    person_instances = [Person(person["name"], person["age"])
+                        for person in people]
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        if person.get("husband"):
+            Person.people[person["name"]
+                          ].husband = Person.people[person["husband"]]
```
SerhiiMar, line 19: flake8 said that line was too long so i fitted it on 2 lines too.

PR #1097, opened 2024-03-18
Solution for task person-class

oleksandra-shershen:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(person_data_list: list) -> list:
+    person_objects = []
+
+    for person_data in person_data_list:
+        person_obj = Person(person_data["name"], person_data["age"])
+        person_objects.append(person_obj)
```
SerhiiMar, line 15: It would be more clear to use use list comprehension

app/main.py
```diff
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
```
SerhiiMar, line 7: Using `Person` here is not best practice

SerhiiMar (changes requested):
Fix the code

app/main.py
oleksandra-shershen: Done

app/main.py
oleksandra-shershen: Done

SerhiiMar (approved):
Looks good!

app/main.py
```diff
@@ -1,8 +1,25 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```

YuriiKindrat (changes requested):

app/main.py
oleksandra-shershen: Done

Oleksl888 (approved):
Well done!

PR #1098, opened 2024-03-18
'Solution'

Vik-By:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_person = [Person(person["name"], person["age"]) for person in people]
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```
Dimosphen1, line 14: You could simplify the check using `get`
```suggestion
        if person.get("wife"):
```
And save value to a variable for further use using the walrus operator
```suggestion
        if wife_name := person.get("wife"):
```

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+            his_wife = Person.people[person["wife"]]
+            Person.people[person["name"]].wife = his_wife
+        if "husband" in person and person["husband"] is not None:
+            her_husband = Person.people[person["husband"]]
+            Person.people[person["name"]].husband = her_husband
```
Dimosphen1, line 19: You could retrieve the `Person.people[person["name"]]` at the beginning of the loop and write it to a variable for further use to avoid code duplication here and on line 16

Dimosphen1 (changes requested):
Several chnages were requested

app/main.py
Vik-By: image

Vik-By: In boolean we have True, but have't Name.

Dimosphen1 (approved):
LGTM, well done!

PR #1099, opened 2024-03-19
Solution

Nyxan:


y-kondrashova (approved):
LGTM👍

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_data in people:
+        name = person_data["name"]
```
Oleksl888, line 13: No need to store name and age in variables since you are not going to reuse them anywhere else. Thus you can simplify this 3 lines of code into one

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        age = person_data["age"]
+        person = Person(name, age)
+        if "wife" in person_data and person_data["wife"] is not None:
```
Oleksl888, line 16: You can simplify this statement using .get method

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+            wife_name = person_data["wife"]
+            wife = Person.people.get(wife_name)
+            person.wife = wife
```
Oleksl888, line 19: This can be moved under if condition since we don't need to assign a wife if it doesnt exist.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+            if wife is not None:
+                wife.husband = person
+        elif "husband" in person_data and person_data["husband"] is not None:
+            husband_name = person_data["husband"]
+            husband = Person.people.get(husband_name)
+            person.husband = husband
+            if husband is not None:
+                husband.wife = person
```
Oleksl888, line 27: This basically duplicates block of code above but with different parameters. What can you do here to follow DRY principle?

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    person_list = []
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person = Person(name, age)
+        if "wife" in person_data and person_data["wife"] is not None:
+            wife_name = person_data["wife"]
+            wife = Person.people.get(wife_name)
+            person.wife = wife
+            if wife is not None:
```
Oleksl888, line 20: we don't need to explicitly check if this variable is None

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_data in people:
+        person = Person(person_data["name"], person_data["age"])
+        if "wife" in person_data and person_data["wife"]:
```
Oleksl888, line 14: you can use .get method to make it more simple

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+            person.wife = Person.people.get(person_data["wife"])
+        elif "husband" in person_data and person_data["husband"]:
+            person.husband = Person.people.get(person_data["husband"])
+            if person.husband:
+                person.husband.wife = person
```
Oleksl888, line 19: You could do without second if :
person.husband =  Person.people.get(person_data["husband"])
husband.wife = person

Oleksl888 (approved):
Good job!

PR #1103, opened 2024-03-19
'Solution'

dkotkod:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
+                self.wife = spouse
+                spouse.husband = self
+            else:
+                self.husband = spouse
+                spouse.wife = self
+
+
+def create_person_list(people_list: list) -> list:
+    person_list = []
+    for person_data in people_list:
```
oleksandra-shershen, line 22: Use list comprehensions here

app/main.py
```diff
@@ -1,8 +1,30 @@
...
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    def set_spouse(self, spouse_name: str) -> None:
```
oleksandra-shershen, line 9: I think, you can implement task without additional methods. But if it was more convenient for you to realize or understand the task, then let it remain as it is.

oleksandra-shershen (changes requested):
Note the remark about the list.
The rest looks really good

app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
YuriiKindrat, line 7: use class name for accessing class attrs

YuriiKindrat (changes requested):

Oleksl888 (approved):
Good stuff!

PR #1104, opened 2024-03-19

bwaah1:


app/main.py
```diff
@@ -1,8 +1,33 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.wife = None
+        self.husband = None
```
semermaks, line 8: a person object should not have the attribute "wife and husband "

semermaks (changes requested):
Check and complite tests

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+    ]
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = Person.people[
+                person["wife"]
+            ]
+            Person.people[person["wife"]].husband = Person.people[
+                person["name"]
+            ]
```
YuriiKindrat, line 23: will sript work without this ?

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+        if person.get("husband"):
+            Person.people[person["name"]].wife = Person.people[
+                person["husband"]
+            ]
+            Person.people[person["husband"]].husband = Person.people[
+                person["name"]
+            ]
```
YuriiKindrat, line 30: same is for this one

YuriiKindrat (changes requested):

PR #1105, opened 2024-03-19
Solution

messedxup:


yossosh (approved):
pretty good!

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_people = [
+        Person(name=human["name"], age=human["age"])
+        for human in people
```
Oleksl888, line 13: This can fit on one line easy

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+    ]
+    for human in people:
+        if human.get("wife"):
+            Person.people[human.get("name")].wife = (
+                Person.people[human.get("wife")]
+            )
+        if human.get("husband"):
+            Person.people[human.get("name")].husband = (
+                Person.people[human.get("husband")]
```
Oleksl888, line 22: same as above

Oleksl888 (changes requested):

app/main.py
messedxup: in both cases flake8 gives an error that the string length is too long

Oleksl888 (approved):
Good job!

PR #1107, opened 2024-03-19
'Solution'

yossosh:


Chub-andrew (approved):
Very good solution.

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+                person.husband = spouse
+                spouse.wife = person
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_data in people:
+        name = person_data["name"]
```
Oleksl888, line 24: No need to declare variables since you will use name and age only once

app/main.py
```diff
@@ -1,8 +1,32 @@
...
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
+
+    @classmethod
```
Oleksl888, line 9: You don't really need classmethod here. You can set attributes of wife and husband simply by accessing Person.people["name"].wife | husband

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+    person_list = []
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person = Person(name, age)
+        person_list.append(person)
+        spouse_key = "wife" if "wife" in person_data else "husband"
+        spouse_name = person_data[spouse_key]
+        if spouse_name:
+            Person.add_spouse(person, spouse_name)
```
Oleksl888, line 31: all this logic is overcomplicated. You can follow this to use less if statements:
if wife in person data -> set person wife to an object that corresponds to wife name + set wifes husband as current person.
repeat same logic for husband.

Oleksl888 (changes requested):
please change

app/main.py
```diff
@@ -1,8 +1,30 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
Oleksl888, line 7: In this case it is better to access people attrbibute through 'Person' class like so: Person.people["name"].
This makes access faster because if you use self.__class__.people[name] construct, python will first looks up the class  of the instance object (self), and then will access the people attribute in that class (that's an extra step). Otherwise it will address the class directly.

Oleksl888 (approved):
Well done!

PR #1108, opened 2024-03-19
Solution

VladyslavKsh:


Gilgumish (approved):
GJ

app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
```
YuriiKindrat, line 7: In this case, it is better to refer to the class attribute through the class name, and not like this

YuriiKindrat: same is for other code

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+    for data in people:
+        person = Person(data["name"], data["age"])
+        if "wife" in data and data["wife"] is not None:
+            if data["wife"] in person.__class__.people:
+                person.wife = person.__class__.people[data["wife"]]
+                person.wife.husband = person
+        elif "husband" in data and data["husband"] is not None:
+            if data["husband"] in person.__class__.people:
+                person.husband = person.__class__.people[data["husband"]]
+                person.husband.wife = person
```
YuriiKindrat, line 21: u have pretty similar parts here, could u optimize this ?
hint: u can iterate through the ['husband', 'wife'] list and use str as keys

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,8 +1,24 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    lover_list = ["husband", "wife"]
+    for data in people:
+        person = Person(data["name"], data["age"])
+        for elem in lover_list:
```
Oleksl888, line 15: spouse will be a better name to call that variable - elem is too general

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+            if elem in data and data[elem] is not None:
```
Oleksl888, line 16: You could use walrus operator to make code more readable. Check it out here: [](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions)

`            if name:= data.get(elem):
                if elem == "wife" and name in Person.people:
                    person.wife = Person.people[name]
                    Person.people[name].husband = person`

PR #1114, opened 2024-03-21
solution

RomanDanyl:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result_list = []
+    for person_data in people:
+        Person.people[person_data["name"]]\
+            = Person(person_data["name"], person_data["age"])
```
Dimosphen1, line 14: There is no need to perform this explicit operation since it is done already on line 7 during the class initialization
```suggestion
        result_list.append(Person(person_data["name"], person_data["age"])
```
Besides that, do not use for loop here, perform list comprehension instead

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+    for person_data in people:
+        person = Person.people[person_data["name"]]
+        if "wife" in person_data and person_data["wife"]:
+            person.wife = Person.people.get(person_data["wife"])
```
Dimosphen1, line 18: You could simplify the condition by using get and walrus operator
```suggestion
        if wife_name := person_data.get("wife"):
            person.wife = Person.people.get(wife_name)
```
Apply this change to the husband condition as well

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1125, opened 2024-03-25
Solution

Ruxxell54:


app/main.py
```diff
@@ -1,8 +1,33 @@
...
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.wife = None
+        self.husband = None
+        self.__class__.people[name] = self
```
YuriiKindrat, line 9: use class name to access class attrs

app/main.py
```diff
@@ -1,8 +1,33 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.wife = None
+        self.husband = None
```
YuriiKindrat, line 8: why do u need this ?

app/main.py
```diff
@@ -1,8 +1,33 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person_data in people:
+        person = Person(person_data["name"], person_data["age"])
+        person_list.append(person)
```
YuriiKindrat, line 17: could u do this with list comprehension ?

YuriiKindrat (changes requested):
all tests should pass

app/main.py
```diff
@@ -1,8 +1,31 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person_data in people:
+        person_list = [Person(person_data["name"],
+                              person_data["age"]) for person_data in people]
+
+    for person_data in people:
```
Oleksl888, line 17: Doing two for loops and using 4 conditional statements is not efficient. Your logic should be like so:
1. Iterate through data with people info.
2. Create a Person on next iteration.
3. if "wife" in data and a Person with wife name already in Person.people list -> set wife and husband attributes.
4. Same check for "husband"
5.  Add Person to the list.
6. return list

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+
+        person = Person(person_data["name"], person_data["age"])
+
+        if "wife" in person_data:
+            wife_name = person_data["wife"]
+            if wife_name in Person.people:
+                person.wife = Person.people[wife_name]
+                person.wife.husband = person
+            person_data["wife"] = None
```
Oleksl888, line 22: why do you need to set this to None? Same on line 29.

Oleksl888 (changes requested):

Oleksl888 (approved):
Nice!

PR #1129, opened 2024-03-27
Func_v1.0

YegorVolkov:
PyTest виводить:
pytest
В той час як ВСЕ працює
Я додав до файлу Self-Check із дружнім до сприйняття виводом
Передперегляд виводу:
task1
Що не так?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,68 @@
 class Person:
-    # write your code here
-    pass
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
+
+    people = {}
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_list = []
+    for human in people:
+        for person in Person.people:
+            if person is human["name"]:
+                new_list.append(Person.people[person])
+        if "wife" in human and human["wife"] is not None:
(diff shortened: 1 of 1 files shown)


NazarSheremet (changes requested):
The issue is that at the moment of iterating over the object you're trying to reference, it might not exist yet. You need to iterate once, creating a list without husband/wife, and then iterate again, adding references to them.

YegorVolkov:
@NazarSheremet Since now I understand the planned idea of the Task - here is the solution to comply with the task requirements

Meanwhile, please take a look on my reworked initial code with complying to what was written in the readme, and your notes:

```
class Person:
    people = {}

    def __init__(self, name: str, age: int) -> None:
        self.name = name
        self.age = age
        Person.people[self.name] = self
        
        
def create_person_list_v1(people: list) -> list:
    n_list = [
        Person.people[human["name"]] for human in people
        if human["name"] in Person.people
    ]

    for human in people:
        person = Person.people[human["name"]]
        if "wife" in human and human["wife"] is not None:
            person.wife = Person.people[human["wife"]]
        if "husband" in human and human["husband"] is not None:
            person.husband = Person.people[human["husband"]]

    return n_list
```

NazarSheremet (approved):
Good job!

PR #1138, opened 2024-04-03
'Solution'

March1205:


vasylhnatiuk (changes requested):
fix tests

March1205:
I don't see how to fix this AssertionError: Person with 'name' Joey should not have attribute wife. I've tried many ways and even tried to assign attribute no wife only to Joey but still nothing. So I would really like to  get some advice about this error

vasylhnatiuk (changes requested):
if you have questions please write them in chat or visit  q&a

PR #1153, opened 2024-04-15
Solution

Jejemira:


app/main.py
```diff
@@ -1,8 +1,38 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name : str, age : int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = Person
 
 
 def create_person_list(people: list) -> list:
```
EdAlekseiev, line 10: To streamline the algorithm, consider simplifying it. You can achieve the desired outcome using just two loops. Begin by creating instances for people within the first loop, leaving the wife and husband fields unfilled initially. Then, in the second loop, assign these fields accordingly.

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1161, opened 2024-04-22
triukhan solution

triukhan:


app/main.py
```diff
@@ -1,8 +1,39 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int | None) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```

app/main.py
```diff
@@ -1,8 +1,39 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int | None) -> None:
```
LiudmylaKulinchenko, line 4: Looks for me that age should not be an optional attribute

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+        return person_instance
+
+
+def create_person_list(people_data: list) -> list:
+    person_list = []
+    for person_data in people_data:
+        person_instance = Person.create_or_get(
+            person_data["name"],
+            person_data.get("age")
+        )
```
LiudmylaKulinchenko, line 25: You can just create a list of Person instances using list comprehension and then iterate by them to add husband, wife attributes

app/main.py
```diff
@@ -1,8 +1,39 @@
...
-    # write your code here
-    pass
+    @classmethod
+    def create_or_get(cls, name: str, age: int = None) -> "Person":
+        if name in cls.people:
+            person_instance = cls.people[name]
+            person_instance.age = age
+        else:
+            person_instance = cls(name, age)
+        return person_instance
```
LiudmylaKulinchenko, line 16: Are you sure we need that check in this task? Try to rewrite without this function to simplify your code

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 🚀

PR #1169, opened 2024-04-29
add solution

midetc:


app/main.py
```diff
@@ -1,8 +1,33 @@
...
+    ]
+    for person, object_person in zip(people, objects_people):
+        if "wife" in person and person.get("wife") is not None:
+            object_person.wife = Person.people.get(person["wife"])
+        elif "husband" in person and person.get("husband") is not None:
+            object_person.husband = Person.people.get(person["husband"])
+    return objects_people
+
+
+people = [
```
vasylhnatiuk, line 29: you don't need it, delete pls

vasylhnatiuk (changes requested):

app/main.py
midetc: delete people?

PR #1170, opened 2024-04-29
add solution

TamaraBuchak:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    # write your code here
-    pass
+    person_list = [
+        Person(person["name"], person["age"])
+        for person in people
+    ]
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = \
+                Person.people[person["wife"]]
```
EdAlekseiev, line 18: Use () instead of `\`. 
```suggestion
            Person.people[person["name"]].wife = (
                Person.people[person["wife"]]
            )
```

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        if person.get("husband"):
+            Person.people[person["name"]].husband = \
+                Person.people[person["husband"]]
```
EdAlekseiev, line 21: The same comment as above

EdAlekseiev (changes requested):

PR #1183, opened 2024-05-15
solution to person class

AlexRozv:
implemented Person class with people dict attribute
implemented create_person_list method

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_dict in people:
+        new_person = Person(person_dict.get("name"), person_dict.get("age"))
+        if "wife" in person_dict and person_dict.get("wife"):
+            new_person.wife = person_dict.get("wife")
+        elif "husband" in person_dict and person_dict.get("husband"):
+            new_person.husband = person_dict.get("husband")
+        person_list.append(new_person)
+    for person in person_list:
+        if hasattr(person, "wife"):
+            person.wife = Person.people.get(person.wife)
+        elif hasattr(person, "husband"):
+            person.husband = Person.people.get(person.husband)
+    return person_list


app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```

app/main.py
```diff
@@ -1,8 +1,21 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        person_list.append(Person(person.get("name"), person.get("age")))
```
LiudmylaKulinchenko, line 13: use list comprehension here

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    for person in people:
+        if "wife" in person and person.get("wife"):
+            husband = Person.people.get(person.get("name"))
```
LiudmylaKulinchenko, line 16: Уou should access the attribute by key without using a method:
1. What happens if Person.people.**get**(person.get("name")) doesn't get a person? We move to the next line `husband.wife` and get an error that None has no attribute wife
2. If we use person.**get**("name"), and it gives us None, then the same problem as above ☝
So if we don't want the result to be None, there is no need to use get method

LiudmylaKulinchenko: Fix everywhere (besides conditions, of course)

LiudmylaKulinchenko (changes requested):

app/main.py
AlexRozv: fixed, but wondering what is the best general practice?
to use a .get() or access by [key]

LiudmylaKulinchenko: Depends on case. If your logic suggests that the value could be None, you should use get
In this case, your code will give unexpected errors like `AttributeError None has no attribute wife` instead of `KeyError <name> is not in person dictionary keys`, something like this

app/main.py
```diff
@@ -1,8 +1,19 @@
...
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    for person in people:
+        if "wife" in person and person["wife"]:
+            husband = Person.people[person["name"]]
+            husband.wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"]:
+            wife = Person.people[person["name"]]
+            wife.husband = Person.people[person["husband"]]
```
LiudmylaKulinchenko, line 18: `Fix everywhere (besides conditions, of course)` -- you've missed the last part 🤫
You should keep method get in conditions

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice job ☀☀

PR #1184, opened 2024-05-16
Solution

MaksymProtsak:


app/main.py
```diff
@@ -1,8 +1,29 @@
...
+
+def create_person_list(p_list: list) -> list:
+    return_list = []
+
+    for i, prs in enumerate(p_list):
+        return_list.append(Person(prs["name"], prs["age"]))
+        if "wife" in prs and prs["wife"] is not None:
+            return_list[i].wife = prs["wife"]
+        if "husband" in prs and prs["husband"] is not None:
+            return_list[i].husband = prs["husband"]
```
EdAlekseiev, line 20: You can simplify your code here with using method `get` of `dict`
```suggestion
        if prs.get("wife"):
            return_list[i].wife = prs["wife"]
        if prs.get("husband"):
            return_list[i].husband = prs["husband"]
```

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+
+    for prs_name in return_list:
+        for maried_prs in return_list:
+            if "wife" in maried_prs.__dict__:
```
EdAlekseiev, line 24: Better use hasattr here
```suggestion
            if hasattr(maried_prs, "wife"):
```

EdAlekseiev (changes requested):

app/main.py
MaksymProtsak: I thinked about it but did non find.

app/main.py
MaksymProtsak: Grate solution!

app/main.py
MaksymProtsak: Done.

app/main.py
MaksymProtsak: Done.

EdAlekseiev (approved):

PR #1190, opened 2024-05-19
py-person-class v1

BornToLivee:


BornToLivee:
i think i accidentally added a couple of extra commits, should i do something with them?

PR #1192, opened 2024-05-20
Solution

VShveda:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    for per in people:
```
DevSpaciX, line 12: Pay attention to naming , imho `per` not a very obvious name (

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
-    # write your code here
-    pass
+
+    for person in people:
+        Person(person["name"], person["age"])
+
+    for person in people:
+        if person.get("wife", None):
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        if "husband" in person and person["husband"]:
```
DevSpaciX, line 18: Just check `if 'husband' in person` that way as you did on line 16 : )
Try always

DevSpaciX (changes requested):

Oleg-s-s (approved):
:+1:

PR #1196, opened 2024-05-21
Solution_V1

BileichukIvan:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
-    # write your code here
-    pass
+    # Creating all Person instances
+    for person_dict in people:
+        Person(person_dict["name"], person_dict["age"])
+
+    # Seting the wife/husband attributes
+    for person_dict in people:
+        person_instance = Person.people[person_dict["name"]]
+        if "wife" in person_dict and person_dict["wife"] is not None:
```
DevSpaciX, line 18: You can check if smth in dict this way -> `if preson_dict.get("wife", None): `
Instead of 2 check as you did 
Same on 20 line

DevSpaciX (changes requested):

Oleg-s-s (approved):
:+1:

PR #1198, opened 2024-05-21
Solution

NikitaNikolaitsev:


app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
```
DevSpaciX, line 12: Try to use list comprehension instead ))

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        person_list.append(Person(person["name"], person["age"]))
+    for person in people:
+        person_instance = Person.people[person["name"]]
+        if "wife" in person and person["wife"]:
```
DevSpaciX, line 16: Also you can check if key in dict using get() . `person.get('wife'):`

DevSpaciX (changes requested):

Oleg-s-s (approved):
👍

PR #1199, opened 2024-05-21
Solution

SrMusienko:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife \
```
DevSpaciX, line 14: All good , but please try to remove this breaking lines ( \ ) . Maybe try to put duplicated code into a viriable  `Person.people[person["name"]]` ??

DevSpaciX (changes requested):

DevSpaciX (approved):
Good ! But Imho `psn` that is not good name of viriable

PR #1210, opened 2024-05-22
solution1

Unlie9:


app/main.py
```diff
@@ -1,8 +1,33 @@
...
+
+def create_person_list(people: list) -> None:
+    person_list = []
+
+    for person_dict in people:
+        name = person_dict["name"]
+        age = person_dict["age"]
+
+        person = Person(name, age)
+        person_list.append(person)
```
Oleg-s-s, line 20: do it in 1 line

app/main.py
```diff
@@ -1,8 +1,33 @@
...
+
+    for person_dict in people:
+        name = person_dict["name"]
+        if "wife" in person_dict and person_dict["wife"] is not None:
```
Oleg-s-s, line 24: you can do this using dict.get() method, same in line 28

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    person_list.extend([Person
+                        (person_dict["name"], person_dict["age"])
+                        for person_dict in people])
+
+    for person_dict in people:
+        name = person_dict["name"]
+        wife_name = person_dict.get("wife")
+        husband_name = person_dict.get("husband")
+
+        if wife_name is not None:
```
DevSpaciX, line 24: If you write  `if wife_name:` it will be the same . As a result you do not need to put `person_dict.get("wife")` into a variable , same with `husband_name`

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        self.name = name
+        self.age = age
+
+        Person.people[name] = self
+
+
+def create_person_list(people: list) -> None:
+    person_list = []
+
+    person_list.extend([Person
```
DevSpaciX, line 15: Read more about list comprehensions , using this will be more clean ))

DevSpaciX (approved):

PR #1211, opened 2024-05-22
sol1

IvanVahanov:


app/main.py
```diff
@@ -1,8 +1,25 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.__class__.people[name] = self
```
Oleg-s-s, line 7: No need to do so complicated, you can simply call `Person` class here)

app/main.py
```diff
@@ -1,8 +1,25 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: dict) -> list:
+    for per in people:
+        Person(per["name"], per["age"])
+
+    for per in people:
```
Oleg-s-s, line 14: per is not a good name, name it somehow more explicit, f.e. if we already have people, name it person, of single_person etc

Oleg-s-s (changes requested):

app/main.py
IvanVahanov: sorry, I not really understand what i should do here

DevSpaciX, line 7: @IvanVahanov You can call here `Person` class . `Person.people[name] = self` is not it ?

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+    for single_person in people:
+        person = Person.people[single_person["name"]]
+        spouse_name = single_person.get("wife") or single_person.get("husband")
+        if spouse_name:
+            spouse = Person.people[spouse_name]
+            if "wife" in single_person:
+                person.wife = spouse
+            else:
+                person.husband = spouse
+            person.spouse = spouse
```
DevSpaciX, line 23: You can simply create 2 separeted checks via `get()` method . First check for wife and second fot husband . 


That way your code will be more clean

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1214, opened 2024-05-22
'Solution'

Oleh323:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_instances = [Person(person["name"], person["age"])
+                        for person in people]
+    for person in people:
+        person_instance = Person.people[person["name"]]
+        if "wife" in person and person["wife"]:
```
DevSpaciX, line 15: You can more simply check if some key in dict using `people.get()` instead of 2 checks

DevSpaciX (changes requested):

app/main.py
Oleh323: Yes, thanks!

DevSpaciX (approved):
Nice

PR #1215, opened 2024-05-23
Solution

Irina17191:


app/main.py
```diff
@@ -1,8 +1,24 @@
 class Person:
-    # write your code here
-    pass
+
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
Oleg-s-s, line 8: You should access class attribute through class name not self

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        person = Person(one["name"], one["age"])
+        if ("wife" in one and one["wife"] is not None
+                and one["wife"] in Person.people):
+            person.wife = Person.people[one["wife"]]
+            person.wife.husband = person
+        elif ("husband" in one and one["husband"] is not None
+              and one["husband"] in Person.people):
+            person.husband = Person.people[one["husband"]]
+            person.husband.wife = person
+        person_list.append(person)
```
Oleg-s-s, line 23: No need to do everything at once, you can divide this in two parts, create person_list and later work with instances

app/main.py
```diff
@@ -1,8 +1,24 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for one in people:
+        person = Person(one["name"], one["age"])
+        if ("wife" in one and one["wife"] is not None
+                and one["wife"] in Person.people):
```
Oleg-s-s, line 16: if ("wife" in one and one["wife"] is not None can be simplified using dict.get()
and one["wife"] in Person.people check is redundant cause you created it a line higher.
Same at line 19-20

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
+
+
+def create_person_list(people: dict) -> list:
+    person_list = []
+    for one in people:
+        person = Person(one["name"], one["age"])
+        person_list.append(person)
```
Oleg-s-s, line 15: You can use list comprehension

Oleg-s-s (changes requested):

DevSpaciX (approved):
Good job!

PR #1217, opened 2024-05-23
'Solution'

ghost:


DevSpaciX (changes requested):

DevSpaciX (approved):
Good , but you coud put  `Person.people[person["name"]]` into a variable and reuse it ) But it is not critical )

PR #1221, opened 2024-05-26
Solution v1

YevaYe:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+        if person.get("husband"):
+            Person.people[person["name"]].husband = (
+                Person.people)[person["husband"]]
```
Oleg-s-s, line 18: bad use of `()`
```suggestion
            Person.people[person["name"]].husband = (
                Person.people[person["husband"]]
            )
```

app/main.py
```diff
@@ -1,8 +1,19 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```
Oleg-s-s, line 7: You should access class attribute through class name not self

Oleg-s-s (changes requested):

app/main.py
DevSpaciX: @YevaYe are you forgot about it ?) It should be `Person.people...`

DevSpaciX (changes requested):

DevSpaciX (approved):
Great !

PR #1225, opened 2024-05-28
Solution

VasylRysan:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = [Person(
+        person["name"],
+        person["age"])
+        for person in people]
+
+    for i, person in enumerate(people):
```
vasylhnatiuk, line 16: use only loop with `range `if you don't use `person `

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1238, opened 2024-06-07
Solution for py-person-class

ArtemShakun:


app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list[dict]) -> list:
+    add_person(people_list)
+    set_spouses(people_list)
+    return list(Person.people.values())
```
Dimosphen1, line 13: There is no need to create these two methods, place all logic within create_person_list

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list[dict]) -> list:
```
Dimosphen1, line 10: Indicate that you return a list of `Person` class instances in the return type annotation

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    add_person(people_list)
+    set_spouses(people_list)
+    return list(Person.people.values())
+
+
+def add_person(people_list: list[dict]) -> None:
+    for person in people_list:
+        name = person["name"]
+        age = person["age"]
+        Person(name, age)
```
Dimosphen1, line 20: There is no need to allocate `name` and `age`variables, pass values directly to Person. Besides that, you could make this operation within list comprehension

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+
+
+def set_spouses(people_list: list[dict]) -> None:
+    for person in people_list:
+        name = person["name"]
+        if "wife" in person and person["wife"]:
+            Person.people[name].wife = Person.people[person["wife"]]
```
Dimosphen1, line 27: You could make this code more efficient using walrus operator to not retrieve value by key several times
```suggestion
        person_instance = Person.people[person["name"]]
        if wife_name := person.get("wife"):
            person_instance.wife = Person.people[wife_name]
```
Apply the same change for husband as well

Dimosphen1 (changes requested):

app/main.py
ArtemShakun: fixed

app/main.py
ArtemShakun: fixed

ArtemShakun: fixed

app/main.py
ArtemShakun: fixed

pavlopro (approved):
Good Job!

PR #1252, opened 2024-06-14
done

Ppllupold:


app/main.py
```diff
@@ -1,8 +1,18 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+        Person.people[name] = self
+
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    person_list = [Person(dict["name"], dict["age"]) for dict in people]
+    for dic in people:
```
vsmutok, line 13: you cannot use the word `dict` as a variable name, as this word is already used in python
also `dic ` not clear variable name

vsmutok (changes requested):

PR #1267, opened 2024-06-27
Solution

NazarSerdiuk1:


app/main.py
```diff
@@ -1,8 +1,41 @@
...
+        self.spouse = None
+        self.wife = None
+        self.husband = None
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
```
pavlopro, line 13: Please simplify this method:

There is no need for the `self.spouse` parameter.
Create persons using list comprehension and store them in a variable.
Iterate over people_list and check if a person has a 'wife' or 'husband'. If they do, retrieve the person object and the spouse object from `Person.people` by their names and assign them accordingly.

pavlopro (changes requested):

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        Person(
+            person_dict["name"], person_dict["age"]
+        )
+        for person_dict in people_list
+    ]
+
+    for person_dict in people_list:
+        person = Person.people[person_dict["name"]]
+        if "wife" in person_dict and person_dict["wife"]:
+            person.wife = Person.people[person_dict["wife"]]
```
Dimosphen1, line 21: You could avoid double retrieval of value by key by using `get` and the walrus operator
```suggestion
        if wife_name := person_dict.get("wife"):
            person.wife = Person.people[wife_name]
```

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+            person.wife.husband = person
```
Dimosphen1, line 22: There is no need to perform this operation since `husband` would be handled separately in other if condition. Remove it here and in the other if condition as well

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
+    [
```
pavlopro, line 11: Store it into variable and return it in the end

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        Person(person_dict["name"], person_dict["age"])
```
pavlopro, line 12: Use unpacking instead:  `Person(**person_dict)`

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        for person_dict in people_list
+    ]
+
+    for person_dict in people_list:
+        person = Person.people[person_dict["name"]]
+        if spouse_name := person_dict.get("wife"):
+            person.wife = spouse = Person.people[spouse_name]
+            spouse.husband = person
```
pavlopro, line 20: There is no need to perform this operation since `husband `would be handled separately in other if condition. Remove it here and in the other if condition as well

pavlopro (changes requested):

pavlopro (approved):
GJ

PR #1280, opened 2024-07-06
Implemented solution

dimak20:


MNDonut (approved):
This code looks better and you don't need to add more loops and ifs in case we need to check for other attributes :)

PR #1288, opened 2024-07-11
Solution

OleksiukStepan:
Implement class **Person** and method **create_person_list()**

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
+
+    for person in people:
+        new_person = Person.people[person["name"]]
+        if "wife" in person and person["wife"]:
+            new_person.wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"]:
```
MNDonut, line 19: it's better to use .get() method of dict because it does both things:
1. Checks if dict has this attribute, if not -> returns None or default value that you set (first part of your conditions)
2. If there is such an attribute, returns its value (second part of your conditions)

MNDonut (changes requested):

MNDonut (approved):

PR #1291, opened 2024-07-15
Solution

apmaz:


app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
Dimosphen1, line 7: `people` is a class attribute, call it from class instead of `self`

app/main.py
```diff
@@ -1,8 +1,23 @@
...
-    # write your code here
-    pass
+    person_list = [
+        Person(name=person["name"], age=person["age"]) for person in people
+    ]
+    for person in people:
+        if "wife" in person and person.get("wife") is not None:
+            Person.people.get(person["name"]).wife = (
+                Person.people.get(person["wife"])
+            )
```
Dimosphen1, line 18: You could simplify your code and avoid line breaking by defining a variable at the beginning of a loop  and using the walrus operator to avoid `is not None` evaluation
```suggestion
    for person in people:
        person_instance = Person.people[person["name"]]
        if wife_name := person.get("wife"):
           person_instance.wife = Person.people.get(wife_name)
```
Apply the same change for husband as well

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1292, opened 2024-07-15
completed coding

aaanyyy:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        self.age = age
+        Person.people[name] = self
+
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    for person in people:
+        Person(person["name"], person["age"])
```
NazarSheremet, line 14: Create a list in which, during the first iteration, you add Person objects without the spouse field, and in the second iteration, add them.

NazarSheremet (changes requested):
Need to fix suggestion

YuriiSmolii (approved):
LGTM

PR #1294, opened 2024-07-15
implement py-person-class

dkn3r:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person_data in people:
```
vasylhnatiuk, line 13: use list comprehension here

vasylhnatiuk (changes requested):

app/main.py
dkn3r: pls check my last commit

vasylhnatiuk (approved):

PR #1297, opened 2024-07-16
Solution

DSahalatyi:


app/main.py
```diff
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
```

app/main.py
```diff
@@ -1,8 +1,22 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        person_instance = Person.people[person["name"]]
+
+        if "husband" in person and person["husband"] is not None:
```

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+            person_instance.husband = Person.people[person["husband"]]
+
+        if "wife" in person and person["wife"] is not None:
```
LiudmylaKulinchenko, line 19: same here

LiudmylaKulinchenko (changes requested):

app/main.py
DSahalatyi: Hi, @LiudmylaKulinchenko!
Could you please elaborate on this one?

MNDonut (approved):

PR #1298, opened 2024-07-16
solution

HalynaRohatska:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
-    pass
+
+    for person in people:
+        Person(person["name"], person["age"])
+
+    for person in people:
+        person_example = Person.people[person["name"]]
+        if person.get("wife"):
+            person_example.wife = Person.people[person["wife"]]
+        elif person.get("husband"):
```
oliasymak24, line 23: What will happend if person.get("husband") will return None?

oliasymak24 (changes requested):
Pay attention to my comments.

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+        self.age = age
+
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
```
MNDonut, line 17: Note that this is a list that you must return, so you can use list comprehension and return it at the end instead of using .values()

MNDonut (changes requested):
One performance improvement required

MNDonut (approved):

PR #1300, opened 2024-07-19
Solution

Atikiho:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = [Person(people[i]["name"], people[i]["age"])
+               for i in range(len(people))]
```
AnyoneClown, line 13: For future reference, the correct plural form is `people`. You should call it `people_list`, for  example

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+    names = {p.name: p for p in persons}
+
+    for i, someone in enumerate(people):
+        if someone.get("wife") is not None:
+            persons[i].wife = names[someone["wife"]]
+        if someone.get("husband") is not None:
+            persons[i].husband = names[someone["husband"]]
+    return persons
```
AnyoneClown, line 22: Try to refactor it in one iteration

AnyoneClown (changes requested):

PR #1301, opened 2024-07-19
solution

georgiy-nezbritskiy:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
+            Person.people[person["name"]].wife = (
+                Person.people)[person["wife"]]
+            continue
```
LiudmylaKulinchenko, line 17: fix formatting
[use this checklist](https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md)

 `braces must be located in one of two options: open and start with the text or have a line break between the text`

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        if "husband" in person and person["husband"] is not None:
+            Person.people[person["name"]].husband = (
+                Person.people)[person["husband"]]
```
LiudmylaKulinchenko, line 20: same here about formatting

app/main.py
LiudmylaKulinchenko, line 17: you anyway won't go in the next condition, so actually not necessary to add a continue here. I understand your logic, but it's not a common practice

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
```
LiudmylaKulinchenko, line 14: You can use `get` method instead of this condition

app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> "Person":
+        self.name = name
+        self.age = age
+        self.people[name] = self
```

app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> "Person":
```
LiudmylaKulinchenko, line 4: You don't return anything in this function, fix annotations

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
+    for person in people:
```
LiudmylaKulinchenko, line 13: You have two same loops now. Let's avoid that. You can create people list using list comprehension from the start, and then return it after updating objects in it

LiudmylaKulinchenko (changes requested):

app/main.py
georgiy-nezbritskiy: if you want I can remove it. but. without "continue" python will go to if statement and check the condition. this will be an unnecessary and avoidable step

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    for person in people:
+        persons = Person.people
+        if person.get("wife") is not None:
```
LiudmylaKulinchenko, line 14: You can remove `is not None` part and it will cover cases where wife is an empty string, for example, which is also helpfull in this case

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🔥

PR #1304, opened 2024-07-20
solution

BastovOleksandr:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
+
+def find_partner(status: str, people: list) -> None:
+    for dict_person in people:
+        if status in dict_person and dict_person[status] is not None:
+            for name, person_link in Person.people.items():
+                if dict_person[status] == name:
+                    if status == "husband":
+                        person_link.wife \
+                            = Person.people[dict_person["name"]]
+                    else:
```
LiudmylaKulinchenko, line 25: avoid using `\` as a line breaker. You can create a new variable or use parentheses

app/main.py
```diff
@@ -1,8 +1,28 @@
...
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    find_partner("husband", people)
+    find_partner("wife", people)
+    return person_list
+
+
+def find_partner(status: str, people: list) -> None:
+    for dict_person in people:
+        if status in dict_person and dict_person[status] is not None:
```
LiudmylaKulinchenko, line 19: You can use `get` method instead of this condition here

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+            for name, person_link in Person.people.items():
+                if dict_person[status] == name:
+                    if status == "husband":
```
LiudmylaKulinchenko, line 22: You can make it one condition, use `and` instead of nesting them

LiudmylaKulinchenko, line 22: Why don't u just iterate through person in people list?

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    find_partner("husband", people)
+    find_partner("wife", people)
```
LiudmylaKulinchenko, line 13: You can avoid creating a new function in this case. It will help you avoid iterating several time throgh the same objects

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for dict_person in people:
+        if dict_person.get("husband"):
+            Person.people[dict_person["husband"]].wife = (
+                Person.people)[dict_person["name"]]
+        elif dict_person.get("wife"):
+            Person.people[dict_person["wife"]].husband = (
+                Person.people)[dict_person["name"]]
+        continue
```
MNDonut, line 20: why do you need continue? Loop automatically iterates through each object in the collection. Continue is used when you need to skip some object like
```
for person in people:
    if person['is_married']:
        continue

```

MNDonut (changes requested):

app/main.py
BastovOleksandr: > why do you need continue? Loop automatically iterates through each object in the collection. Continue is used when you need to skip some object like
> 
> ```
> for person in people:
>     if person['is_married']:
>         continue
> ```

most likely this operator simply overlooked it when reworking the previous solution :(

MNDonut (approved):

PR #1305, opened 2024-07-20
Resolve

ihorutkin:


app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
+
+    print(Person.people)
```
MNDonut, line 18: you can remove this redundant print

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+
+    for person in people:
+        instance = Person.people[person["name"]]
+
+        if "wife" in person and person["wife"]:
+            instance.wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"]:
+            instance.husband = Person.people[person["husband"]]
```
MNDonut, line 26: instead of 2 conditions you can use .get() method of dict

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        result.append(Person.people[person["name"]])
+
+    return result
+
+
+people = [
+    {"name": "Ross", "age": 30, "wife": "Rachel"},
+    {"name": "Joey", "age": 29, "wife": None},
+    {"name": "Rachel", "age": 28, "husband": "Ross"}
+]
```
MNDonut, line 40: you don't need this testing data

app/main.py
```diff
@@ -1,8 +1,40 @@
...
+        if "wife" in person and person["wife"]:
+            instance.wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"]:
+            instance.husband = Person.people[person["husband"]]
+
+    result = []
+
+    for person in people:
+        result.append(Person.people[person["name"]])
+
```
MNDonut, line 32: it's better to create a list of persons in the first loop, so you don't need to loop through all persons twice

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for person in people:
+        Person(person["name"], person["age"])
+        result.append(Person.people[person["name"]])
```
MNDonut, line 18: In this implementation it would be better to assign person to some variable and append this variable to result instead of fetching the same person from people class attribute
However it's possible to simplify it using list comprehension, let's do it :)

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌷

PR #1306, opened 2024-07-21
Everything works fine but now I hate this series =/

Narberal90:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for pers in people:
+        if pers["name"] not in Person.people:
+            Person(pers["name"], pers["age"])
```
MNDonut, line 13: As this function is called only once, it's impossible there are some persons, so you can just create Person instances. Also you can use list comprehension here

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+    for pers in people:
+        person = Person.people[pers["name"]]
+        if "wife" in pers and pers["wife"] is not None:
+            person.wife = Person.people[pers["wife"]]
+        if "husband" in pers and pers["husband"] is not None:
+            person.husband = Person.people[pers["husband"]]
```
MNDonut, line 20: instead of making 2 conditions, you can use .get() method of dict

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    [Person(pers["name"], pers["age"]) for pers in people]
```
MNDonut, line 11: We don't do list comprehensions in this way.
Note this requirement: Write function create_person_list, this function takes list people and **return list with Person instances** instead of dicts.
So you can assign this comprehension to variable and remove it at the end instead of generating a new list comprehension

MNDonut (changes requested):

MNDonut (approved):

PR #1307, opened 2024-07-21
Solution

mnezbrytska:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for one_person in people:
+        person_list.append(
+            Person(name=one_person["name"], age=one_person["age"])
+        )
```
MNDonut, line 17: When we have only append operation and some if statements, then we are free to use list comprehension :)

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    for one_person in people:
+        if "wife" in one_person and one_person["wife"] is not None:
+            Person.people[one_person["name"]].wife = \
+                Person.people[one_person["wife"]]
+        if "husband" in one_person and one_person["husband"] is not None:
+            Person.people[one_person["name"]].husband = \
+                Person.people[one_person["husband"]]
```
MNDonut, line 24: You can use .get() method of dict that does both checks at the same time ;)
Also it's better to use parenthesis as a line breaker instead of \ if possible

MNDonut (changes requested):

MNDonut (approved):
clean code!

PR #1309, opened 2024-07-22
'Solution'

dartomOOv:


app/main.py
```diff
@@ -1,8 +1,27 @@
 class Person:
-    # write your code here
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
     pass
```
MNDonut, line 7: you can remove this pass

MNDonut, line 7: this requirement isn't met:
This class also should have a class attribute people, it is a dict that stores Person instances by their name. The __init__ method should add elements to this attribute.

app/main.py
```diff
@@ -1,8 +1,27 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+    Person.people = {person.name: person for person in person_list}
```
MNDonut, line 12: after you fix the requirement above, you can remove this assignment

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+
+    for i in range(len(person_list)):
```
MNDonut, line 14: instead of raw indexes you can just loop in this way:
`for person in person_list`
In this way you instantly get user dict object

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        if list(people[i].values())[2] is not None:
+            for person in person_list:
+                if people[i][list(people[i].keys())[2]] == person.name:
+
+                    if list(people[i].keys())[2] == "wife":
+                        person_list[i].wife = person
+                    else:
+                        person_list[i].husband = person
+
+                    break
```
MNDonut, line 25: you can simplify it after the comment above will be fixed ;)

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+                Person.people)[person["husband"]]
+
+    return person_list
+
+
+people = [
+    {"name": "Ross", "age": 30, "wife": "Rachel"},
+    {"name": "Joey", "age": 29, "wife": None},
+    {"name": "Rachel", "age": 28, "husband": "Ross"}
+]
```
MNDonut, line 29: remove testing data

app/main.py
```diff
@@ -1,8 +1,29 @@
...
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if "wife" in person and person["wife"] is not None:
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+
+        if "husband" in person and person["husband"] is not None:
+            Person.people[person["name"]].husband = (
+                Person.people)[person["husband"]]
```
MNDonut, line 20: You can use .get() method of dict instead of using two conditions

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    # write your code here
-    pass
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if person.get("wife") is not None:
+            Person.people[person["name"]].wife = Person.people[person["wife"]]
+
+        if person.get("husband") is not None:
+            Person.people[person["name"]].husband = (
```
MNDonut, line 19: You don't need to check is not None.
person.get(...) is enough. Python understands that you need something, that isn't empty, like empty list, None, False, etc

MNDonut (changes requested):

MNDonut (approved):

PR #1311, opened 2024-07-22
create Person class, create_person_list func

vladislav-tsybuliak1:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
```
MNDonut, line 12: Note that this is a list you must return at the end, so let's use here list comprehension and return it instead of using .values() at the end

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+    for person in people:
+        if "wife" in person and person["wife"]:
+            Person.people[person["name"]].wife = \
+                Person.people[person["wife"]]
+        elif "husband" in person and person["husband"]:
+            Person.people[person["name"]].husband = \
+                Person.people[person["husband"]]
```
MNDonut, line 20: 1. When you need to check this both conditions, you can use .get() method of dict that does the same thing
2. We don't use \ as a line breaker. It's better to use parenthesis
3. If you don't want to use \ and () you can, for example, move fetching person by name in a separate variable and reuse it because it's used in both if statements

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Looks good 🔥

PR #1314, opened 2024-07-23
implement py-person-class

ValeriiaKyr:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_person = []
+    names = {}
```
MNDonut, line 13: you don't need names here because you have people in Person class

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+    for human in people:
+        person = Person(human["name"], human["age"])
+        list_person.append(person)
+        names[human["name"]] = person
```
MNDonut, line 17: let's use here list comprehension

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+
+    for human in people:
+        person = names[human["name"]]
+        if "wife" in human and human["wife"] is not None:
+            if human["wife"] in names:
+                person.wife = names[human["wife"]]
+        if "husband" in human and human["husband"] is not None:
+            if human["husband"] in names:
+                person.husband = names[human["husband"]]
```
MNDonut, line 26: instead of `wife in human and human['wife']..` you can use .get() method of dict
also `human['wife'] will be in people exactly because you add all persons before this loop`

MNDonut (changes requested):

MNDonut (approved):

PR #1315, opened 2024-07-23
Solution for py-person-class

AndriiHamasa:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    res_arr = []
+    for person in people:
+        res_arr.append(Person(person["name"], person["age"]))
```
MNDonut, line 14: let's use list comprehension here

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+    for i, person in enumerate(people):
+        if person.get("wife"):
+            res_arr[i].wife = Person.people[person["wife"]]
+        if person.get("husband"):
+            res_arr[i].husband = Person.people[person["husband"]]
```
MNDonut, line 20: let's iterate through values, not indexes

MNDonut (changes requested):

app/main.py
AndriiHamasa: I don't understand how we can realize it without indexes? Сan you explain more broadly?

MNDonut (approved):

PR #1317, opened 2024-07-23
first_result

yashkunn:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person in people:
+        name = person.get("name")
+        age = person.get("age")
+        person_instance = Person(name, age)
+        person_list.append(person_instance)
```
MNDonut, line 17: let's use list comprehension here

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [
+        Person(person.get("name"),
+               person.get("age"))
```
MNDonut, line 13: doesn't it fit in 1 line in less than 79 characters?

MNDonut: and if not, then parameters should be passed from the next lines after class name

MNDonut (changes requested):

oliasymak24 (approved):
Good job!

PR #1319, opened 2024-07-23
Solution

Alexandr-Politov:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for someone in people:
+        Person(someone["name"], someone["age"])
```
MNDonut, line 13: 1. Let's use list comprehension here
2. Let's save these person and return at the end instead of fetching them from Person.people.values()

MNDonut (changes requested):

MNDonut (approved):
This is exactly what I wanted!

PR #1321, opened 2024-07-23
Solution

Bogdan-Rylov:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_objects = {
+        person["name"]: Person(person["name"], person["age"])
+        for person in people
+    }
```
MNDonut, line 14: 1. Why is it dict, not list that should be returned?
2. You have Person.people dict for this purpose

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = [
+        Person(person["name"], person["age"])
+        for person in people
+    ]
+
+    for person in people:
```
Oleksl888, line 16: Can you do all this in one iteration to optimize performance of your code?

Oleksl888 (changes requested):

app/main.py
Bogdan-Rylov: I'm not sure it's a good idea since the input is a `list` of `dict`, and a `spouse` attribute requires an object. To add a `person_object` to the result list, it's necessary to check if the `person_object` is already created (as a spouse for a previous object). Then, check if the `spouse_object` is already in `Person.people`, and if not, iterate further through the `people` list to find `age` to create it. This would decrease the readability of the code significantly with a slight performance boost. Should I still do it?

app/main.py
```diff
@@ -1,8 +1,27 @@
...
-    # write your code here
-    pass
+    people_list = []
+
+    for person in people:
+        person_object = Person(person["name"], person["age"])
+        people_list.append(person_object)
+        spouse_name = person.get("wife") or person.get("husband")
+
+        if spouse_name and spouse_name in Person.people:
```
Oleksl888, line 18: `spouse_name in Person.people` can be replaced with `Person.people.get(spouse_name)` and it will give you the best performance

Oleksl888 (changes requested):

MNDonut (approved):

PR #1323, opened 2024-07-23
'Solution'

ABaidin:


app/main.py
```diff
@@ -1,8 +1,31 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person = Person(name, age)
```
MNDonut, line 15: You can directly fetch and pass values to Person and make list comprehension

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+
+        if "wife" in person_data and person_data["wife"]:
```
MNDonut, line 17: it's better to use .get() method of dict

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+            wife_name = person_data["wife"]
+            if wife_name in Person.people:
+                person.wife = Person.people[wife_name]
+                Person.people[wife_name].husband = person
+
+            elif "husband" in person_data and person_data["husband"]:
+                husband_name = person_data["husband"]
+                if husband_name in Person.people:
+                    person.husband = Person.people[husband_name]
+                    Person.people[husband_name].wife = person
```
MNDonut, line 27: how will this condition work if the main condition checks `if wife in person_data`. As only 1 parameter can be passed, then this condition will never work

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+    person_list = [
+        Person(person_data["name"], person_data["age"])
+        for person_data in people
+    ]
+
+    for person in people:
+
+        if "wife" in person:
+            if person.get("wife"):
+                Person.people[person.get("name")].wife =\
```
Oleksl888, line 20: Rewrite this so you don't use `\` for breaking lines

Oleksl888 (changes requested):

app/main.py
Oleksl888: not fixed.

Oleksl888 (changes requested):

app/main.py
ABaidin: Sorry, I just woke up :)

app/main.py
```diff
@@ -1,8 +1,30 @@
...
-    pass
+    person_list = [
+        Person(person_data["name"], person_data["age"])
+        for person_data in people
+    ]
+
+    for person in people:
+
+        if "wife" in person:
+            if person.get("wife"):
```
MNDonut, line 19: .get does 2 things:
1. Checks if this key exist. If no -> returns default value(None) and KeyError will not be raised
2. If key exists -> returns its value, this you can simplify these if statements

MNDonut (changes requested):

oliasymak24 (approved):

PR #1324, opened 2024-07-24
'Solution'

Ivan-Shakhman:
Solution of task

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.spouse = None
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for human in people:
+        name = human["name"]
+        age = human["age"]
+        Person(name, age)
+
+    for human in people:
+        person = Person.people[human["name"]]
+        spouse_name = human.get("wife") or human.get("husband")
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,26 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for human in people:
+        name = human["name"]
+        age = human["age"]
+        Person(name, age)
```
MNDonut, line 15: you can make list comprehension and return it instead of using .values()

app/main.py
```diff
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.spouse = None
```
MNDonut, line 7: spouse shouldn't be a class attribute

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+
+    for human in people:
+        person = Person.people[human["name"]]
+        spouse_name = human.get("wife") or human.get("husband")
+        if spouse_name:
+            person.spouse = Person.people[spouse_name]
+            if "wife" in human:
+                person.wife = person.spouse
+            elif "husband" in human:
+                person.husband = person.spouse
```
MNDonut, line 25: you can use .get() instead of these conditions

MNDonut (changes requested):

MNDonut (approved):

PR #1325, opened 2024-07-24
Solution

pashawarganov:


app/main.py
```diff
@@ -1,8 +1,28 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.wife = None
+        self.husband = None
```
MNDonut, line 8: class shouldn't have these attributes

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_data: list) -> list:
+    class_list = []
+    for person_data in people_data:
+        class_list.append(Person(person_data["name"], person_data["age"]))
```
MNDonut, line 15: you can use list comprehension here

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+    for i, person_data in enumerate(people_data):
+        if person_data.get("wife") is not None:
+            class_list[i].wife = Person.people[person_data["wife"]]
+        else:
+            delattr(class_list[i], "wife")
+
+        if person_data.get("husband") is not None:
+            class_list[i].husband = Person.people[person_data["husband"]]
+        else:
+            delattr(class_list[i], "husband")
```
MNDonut, line 26: 1. Let's iterate through objects, not indexes
2. No need to check `is not None` because python considers `None` as `False`
3. Let's don't use delattr at all

MNDonut (changes requested):

MNDonut (approved):

PR #1326, opened 2024-07-24
Solution

mihavryliuk:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
-    # write your code here
-    pass
+        Person.people[name] = self
+
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    for person_dict in people:
+        name = person_dict["name"]
+        age = person_dict["age"]
+        Person(name, age)
```
MNDonut, line 15: you can make it in 1 line if directly pass values + use list comprehension

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+        if spouse_name:
+            person = Person.people[name]
+            spouse = Person.people.get(spouse_name)
+            if spouse:
+                if "wife" in person_dict:
+                    person.wife = spouse
+                elif "husband" in person_dict:
+                    person.husband = spouse
+
+    return list(Person.people.values())
```
MNDonut, line 30: you can save values from the first loop and return them here to not loop through all values again

MNDonut (changes requested):

MNDonut (approved):

PR #1330, opened 2024-07-24
create class person and method create_person_list and return list wit…

egor250330:
…h Person instances instead of dicts

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,27 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int):
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+def create_person_list(people_list):
+    person_instances = []
+    for person_dict in people_list:
+        person = Person(person_dict["name"], person_dict["age"])
+        person_instances.append(person)
+
+    for person_dict in people_list:
+        person = Person.people[person_dict["name"]]
+        if "wife" in person_dict and person_dict["wife"]:
+            spouse = Person.people[person_dict["wife"]]
+            person.wife = spouse
+            spouse.husband = person
+        elif "husband" in person_dict and person_dict["husband"]:
+            spouse = Person.people[person_dict["husband"]]
+            person.husband = spouse
+            spouse.wife = person
+
+    return person_instances


app/main.py
```diff
@@ -1,8 +1,27 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
+    person_instances = []
+    for person_dict in people_list:
+        person = Person(person_dict["name"], person_dict["age"])
+        person_instances.append(person)
```
MNDonut, line 14: you can use list comprehension here

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+
+    for person_dict in people_list:
+        person = Person.people[person_dict["name"]]
+        if "wife" in person_dict and person_dict["wife"]:
+            spouse = Person.people[person_dict["wife"]]
+            person.wife = spouse
+            spouse.husband = person
+        elif "husband" in person_dict and person_dict["husband"]:
```
MNDonut, line 22: you can use get() method of dict instead of these two conditions

MNDonut (changes requested):

app/main.py
```diff
@@ -8,20 +8,19 @@ def __init__(self, name: str, age: int) -> None:
 
 
 def create_person_list(people_list: list) -> list:
-    person_instances = []
-    for person_dict in people_list:
-        person = Person(person_dict["name"], person_dict["age"])
-        person_instances.append(person)
+    person_instances = [Person(person_dict["name"], person_dict["age"])
+                        for person_dict in people_list]
```
MNDonut, line 12: move the body inside list comprehension on separate lines as we do for function parameters, etc

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+
+def create_person_list(people_list: list) -> list:
+    person_instances = [
+        Person(
+            person_dict["name"], person_dict["age"]
+        )
+        for person_dict in people_list
+    ]
+
+    for person_dict in people_list:
```
Oleksl888, line 18: Can you rewrite this so you don't have to iterate twice through the same data?

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #1331, opened 2024-07-25
Task 6

Poznanskyi:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person_dict in people:
+        name = person_dict["name"]
+        age = person_dict["age"]
+        person = Person(name, age)
+        person_list.append(person)
```
MNDonut, line 17: let's use list comprehension here

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+
+    for person_dict in people:
+        person = Person.people[person_dict["name"]]
+
+        if "wife" in person_dict and person_dict["wife"]:
+            person.wife = Person.people[person_dict["wife"]]
+
+        if "husband" in person_dict and person_dict["husband"]:
+            person.husband = Person.people[person_dict["husband"]]
```
MNDonut, line 26: use can use .get() method of dict instead of two conditions

MNDonut (changes requested):

MNDonut (approved):

PR #1344, opened 2024-08-03
Solution

alheimurin-n:


YuriiSmolii (changes requested):
If person's key wife/husband is not None - create_person_list should add attribute wife/husband respectively to its instance. This attribute should be a link to a Person instance with name the same as wife/husband key in person's dict.

alheimurin-n:
@YuriiSmolii could you check please?

NazarSheremet (approved):
Good job

PR #1349, opened 2024-08-07
Solution

bdnnk:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list[Person]:
+
+    for person in people:
+        Person(person["name"], person["age"])
```
Dimosphen1, line 13: You could perform this operation in a list comprehension instead of for loop

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+
+    for person in people:
+        person_instance = Person.people[person["name"]]
+
+        if "wife" in person and person["wife"]:
+            person_instance.wife = Person.people[person["wife"]]
```
Dimosphen1, line 19: You could avoid having to checks with `and` by using `get()`
```suggestion
        if person.get("wife"):
            person_instance.wife = Person.people[person["wife"]]
```
and reduce the need to retrieve value by key twice by using walrus operator
```suggestion
        if wife_name := person.get("wife"):
            person_instance.wife = Person.people[wife_name]
```
Apply similar change for `husband` as well

Dimosphen1 (changes requested):
Several minor changes were requested

pavlopro (approved):
Good Job!

PR #1355, opened 2024-08-11
My solution

lavryniuk-dev:


app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people.update({name: self})
```

LiudmylaKulinchenko, line 7: Using assigning directly by the key, without update, will give you same output here. You should use update when you need to update or merge multiple key-value pairs at once

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great job 😸

PR #1364, opened 2024-08-16
Solution

Anzda:


vasylhnatiuk (changes requested):
you use very hard version of code, try to simplify it, use only two seperate loops and don't use i, j as name for varibles

vasylhnatiuk (approved):

PR #1391, opened 2024-09-08
'Solution'

mshmygel:


AnyoneClown (approved):
LGTM, but u can do some changes

app/main.py
```diff
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
```
AnyoneClown, line 7: You should access class attribute through class name not self

app/main.py
```diff
@@ -1,8 +1,22 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    for person_data in people:
+        Person(person_data["name"], person_data["age"])
```
AnyoneClown, line 13: Use list comprehension here

AnyoneClown (commented):
Fix this

PR #1398, opened 2024-09-13
people

dmytrik:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = []
+    for person in people:
+        people_list.append(Person(person["name"], person["age"]))
```
AnyoneClown, line 13: Use list comprehension here

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+    for i in range(len(people_list)):
+        if people[i].get("wife"):
+            people_list[i].wife = Person.people.get(people[i]["wife"])
+            continue
+        if people[i].get("husband"):
+            people_list[i].husband = Person.people.get(people[i]["husband"])
```
AnyoneClown, line 20: You don't need to do a `range` loop. U can simplify it

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    people_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if person.get("wife"):
+            Person.people.get(person.get("name")).wife = (
+                Person.people.get(person.get("wife")))
```
Nattalli, line 19: bad code style, fix it. Here is an example: https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md#code-style

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+            continue
```
Nattalli, line 20: do you really need that `continue` here?

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        if person.get("husband"):
+            Person.people.get(person.get("name")).husband = (
+                Person.people.get(person.get("husband")))
```
Nattalli, line 23: not readable code, improve it by creating new variables

Nattalli: also fix braces style here

Nattalli (approved):
Good, but please, fix variable name `people_list`, Can it be `people`? Seems better 🌞

PR #1407, opened 2024-09-21
Solution-py-person-class

skyfoxwork:


Mxbely (approved):
we can use just `Person.people[name] = self`, because we just add new key: val to dict, but a think, your solution also good

and this solution bit difficult to read, but i approve it
```
if person.get("wife"):
            (Person.people[person.get("name")].__dict__
             .update({"wife": Person.people[person.get("wife")]}))
```

app/main.py
```diff
@@ -1,8 +1,22 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people.update({self.name: self})
```
AnyoneClown, line 7: Fix this(Comment above)

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if person.get("wife"):
+            (Person.people[person.get("name")].__dict__
+             .update({"wife": Person.people[person.get("wife")]}))
+
+        if person.get("husband"):
+            (Person.people[person.get("name")].__dict__
+             .update({"husband": Person.people[person.get("husband")]}))
```
AnyoneClown, line 20: Try to simplify this code. We don't need to use `update` method for this.

Example:
`Person.people[name].wife = Person.people[wife_name]`

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.name = name
+        self.age = age
+        Person.people.update({self.name: self})
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    person_list = [Person(person["name"], person["age"]) for person in people]
```
AnyoneClown, line 11: Let's use `.get()` method for all cases

AnyoneClown (changes requested):

app/main.py
Nattalli: well use `.get()` method everywhere instead of  getting data through `.[]`

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+    for person in people:
+        if person.get("wife"):
+            Person.people[person.get("name")].wife \
+                = Person.people[person.get("wife")]
```
Nattalli, line 16: also recommend use https://peps.python.org/pep-0572/#exceptional-cases

```suggestion
        if wife := person.get("wife"):
            Person.people[person.get("name")].wife = Person.people[wife]
```

Nattalli (changes requested):

PR #1414, opened 2024-09-25
Solution

VladyslavBon:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    instances = {}
+
+    for person in people:
+        person_instance = Person(person["name"], person["age"])
+        instances[person["name"]] = person_instance
```
AnyoneClown, line 15: Use dict comprehension here

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+
+    for person in people:
+        if "wife" in person and person["wife"]:
```
AnyoneClown, line 18: We can simplify it just using `if person.get("wife")` condition. Check all cases

AnyoneClown: We can also use warlus oprator here(Inside if condition)

https://docs.python.org/3/whatsnew/3.8.html

app/main.py
```diff
@@ -1,8 +1,25 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
```
AnyoneClown, line 10: Add more descriptive `return` type hint

AnyoneClown (changes requested):

AnyoneClown:
Also, be sure that your code passed all tests on github

Nattalli (approved):

PR #1416, opened 2024-09-26
solution

sashasyrota:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        attr = Person(person["name"], person["age"])
+        result.append(attr)
+        Person.people.update({person["name"]: attr})
+
+    for count_wife in range(len(people)):
+        if people[count_wife].get("wife") is not None:
+            for count_husband in range(len(people)):
+                if people[count_wife]["wife"] == people[count_husband]["name"]:
+                    result[count_wife].wife = result[count_husband]
+                    result[count_husband].husband = result[count_wife]
```
LLkaia, line 22: This approach is not good since you are relying on order of lists. Do not use indexes, please

LLkaia (changes requested):

app/main.py
```diff
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: str) -> None:
+        self.name = name
+        self.age = age
+        self.people.update({self.name: self})
```
Dimosphen1, line 7: Call class attribute from class, not from an instance
```suggestion
        Person.people.update({self.name: self})
```

app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = [Person(person["name"], person["age"]) for person in people]
+
+    for name_wife in people:
+        if name_wife.get("wife") is not None:
```
Dimosphen1, line 14: There is no need to check this condition with `is not None`. Besides that, you could allocate a wife`s name in a variable and simplify the internals of the if condition. Take into account, that you should not place arguments on the same line with `]` in case `[` does not have arguments on the same line
```suggestion
        if wife_name := name_wife.get("wife"):
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
-    # write your code here
-    pass
+    result = [Person(person["name"], person["age"]) for person in people]
+
+    for name_person in people:
+        if wife_name := name_person.get("wife"):
+            Person.people[name_person["name"]].wife \
+                = Person.people[wife_name]
+            Person.people[name_person["name"]].wife.husband \
+                = (Person.people)[name_person["name"]]
```
LLkaia, line 18: Do not use \ for new lines since it is bad practice. You can do your lines shorter by writing name_person["wife"] to separate variable. Otherwise, use () to split line

LLkaia (changes requested):

LLkaia (changes requested):

LLkaia (approved):
LGTM

PR #1423, opened 2024-09-28
implement py-person-class

yuliiapyrih:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
```
LiudmylaKulinchenko, line 12: You can create a list here using list comprehension, then iterate through it and return the same list (you create it in the return anyway)

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+
+    for person in people:
+        person_inst = Person.people[person["name"]]
+        if "wife" in person and person["wife"] is not None:
```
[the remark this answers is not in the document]
yuliiapyrih: Done

app/main.py
yuliiapyrih: Done

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        person_inst = Person.people[person["name"]]
```
Nattalli: you make variable name shorter for few characters, suggest you just use full name:
```suggestion
        person_instance = Person.people[person["name"]]
```

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+        if person.get("wife"):
+            person_inst.wife = Person.people[person["wife"]]
```
Nattalli: better use only `.get()` method or only `[]` for taking values

app/main.py
```diff
@@ -1,8 +1,20 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
```
Nattalli:
```suggestion
def create_person_list(people: list[dict]) -> list[Person]:
```

Nattalli (changes requested):

app/main.py
yuliiapyrih: Done

app/main.py
yuliiapyrih: Done

app/main.py
yuliiapyrih: Done

Nattalli (approved):

PR #1435, opened 2024-10-01
solution

YaYaYaroslav:
I use 2 different cycles to avoid a mistake KeyError: 'Rachel'

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    for person in people:
+        Person(person["name"], person["age"])
+
+    for person in people:
+        person_instance = Person.people[person["name"]]
+
+        if "wife" in person and person["wife"]:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,23 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    for person in people:
+        Person(person["name"], person["age"])
+
+    for person in people:
```
Nattalli, line 15: can you use a list? comprehension for creating a new list with `Person` items?

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        person_instance = Person.people[person["name"]]
+
+        if "wife" in person and person["wife"]:
```
Nattalli, line 18: you can avoid double checking with the usage of `.get()` method

Nattalli (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    pass
+
+def create_person_list(people: list[dict]) -> list[Person]:
+
+    [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        person_instance = Person.people[person["name"]]
+
+        if "wife" in person and person.get("wife"):
```
Nattalli, line 17: just to make more beautiful you can use this:
```suggestion
        if wife := person.get("wife"):
```

Nattalli (approved):
good job

PR #1436, opened 2024-10-01
Solution

Artem773:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person_data in people:
+        person_instance = Person(name=person_data["name"],
+                                 age=person_data["age"])
+        person_list.append(person_instance)
```
94nj111, line 19: you can use list comprehension here

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+
+    for person_data in people:
+        person_instance = Person.people[person_data["name"]]
+        if "wife" in person_data and person_data["wife"]:
+            person_instance.wife = Person.people[person_data["wife"]]
+        if "husband" in person_data and person_data["husband"]:
```
94nj111, line 25: you can just use __get()__ instead of that __and__ condition, and also you can use walrus operetor := in __if__ condition, so you don't need to call person_data["wife"] twice, same for husband

94nj111 (approved):

app/main.py
Nattalli: agree, @94nj111 next time please reject if you have comments

Nattalli: list comprehension would be more readable as it's just list creating

app/main.py
Nattalli: good comment, agree!

Nattalli (changes requested):

PR #1438, opened 2024-10-01
solution

VolodymyrSemchysyn:


app/main.py
```diff
@@ -1,8 +1,30 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.wife = None
+        self.husband = None
```
molodsh1y1, line 8: These attributes should not be here

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+    for person_data in people:
+        person = Person.people[person_data["name"]]
+        if "wife" in person_data and person_data["wife"] is not None:
+            person.wife = Person.people[person_data["wife"]]
+        else:
+            del person.wife
+        if "husband" in person_data and person_data["husband"] is not None:
+            person.husband = Person.people[person_data["husband"]]
+        else:
+            del person.husband
```
molodsh1y1, line 28: Better used get(), you can simplify the code to just check if person_data.get("wife") etc...
And you don't need to delete the data

app/main.py
```diff
@@ -1,8 +1,30 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person_data in people:
+        person = Person(person_data["name"], person_data["age"])
+        person_list.append(person)
```
molodsh1y1, line 17: you could use the list comprehension

molodsh1y1 (changes requested):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        Person(person_data.get("name"), person_data.get("age"))
+        for person_data in people
+    ]
+
+    for person_data in people:
+        person = Person.people[person_data.get("name")]
+        if "wife" in person_data and person_data.get("wife") is not None:
+            person.wife = Person.people[person_data.get("wife")]
+        if "husband" in person_data and person_data.get("husband") is not None:
+            person.husband = Person.people[person_data.get("husband")]
```
molodsh1y1, line 21: You can just write if person_data.get(“wife”) and it will still work. If the key doesn't exist it returns None.
And you don't need to check if "husband" is in person_data because you check it with get.

molodsh1y1 (changes requested):
lgtm

app/main.py
```diff
@@ -1,8 +1,23 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
```
Nattalli, line 10: can you specify types of list items?

app/main.py
Nattalli: agree, checking `and person_data.get("husband") is not None` is unnecessary

Nattalli (changes requested):

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    person_list = [
+        Person(person_data.get("name"), person_data.get("age"))
+        for person_data in people
+    ]
 
+    for person_data in people:
+        person = Person.people[person_data.get("name")]
+        if "wife" in person_data and person_data.get("wife"):
```
Nattalli, line 20:
```suggestion
        if wife := person_data.get("wife"):
```

Nattalli (approved):

PR #1444, opened 2024-10-02
Solution

zaietsmo:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for person in people:
+        person_instance = Person(person["name"], person["age"])
+        result.append(person_instance)
```
molodsh1y1, line 14: It would look better if you could rewrite it using list comprehension.

molodsh1y1 (approved):
LGTM

app/main.py
Nattalli: agree, use list comprehension

Nattalli (changes requested):

PR #1447, opened 2024-10-02
Solution

MatvijSafonov:


app/main.py
```diff
@@ -1,8 +1,30 @@
...
+    for person in person_instances:
+        for original_person in people_list:
+            if original_person["name"] == person.name:
+                spouse_key = "wife" if "wife" in original_person else "husband"
+                spouse_name = original_person.get(spouse_key)
+                if spouse_name:
+                    if spouse_key == "wife":
+                        person.wife = Person.people[spouse_name]
+                    else:
+                        person.husband = Person.people[spouse_name]
```
dxrrkwm, line 28: you can use for loop with walrus operator here to make code more readable and simplier

app/main.py
```diff
@@ -1,8 +1,30 @@
...
-    pass
+
+def create_person_list(people_list: list) -> list:
+    person_instances = []
+
+    for person in people_list:
+        name = person["name"]
+        age = person["age"]
+        person_instance = Person(name, age)
+        person_instances.append(person_instance)
```
dxrrkwm, line 17: also you can use list comprehension here so you don't need these instances

dxrrkwm (changes requested):

MatvijSafonov (commented):
is that what you meant ?

dxrrkwm (changes requested):
you don't need to use nested loop here and also you can create list using comprehension, try to do that :)

dxrrkwm (approved):
looks good!

MatvijSafonov:
> looks good!

thx

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
```
Nattalli, line 10: can you specify the types of list items

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    person_instances = [
+        Person(
+            person.get("name"),
+            person.get("age")
+        ) for person in people_list
+    ]
+
+    people_dict = {person.get("name"): person for person in people_list}
```
Nattalli, line 18: you can use only one comprehension, two is not optimal

Nattalli (changes requested):

app/main.py
```diff
@@ -1,8 +1,19 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
```
Nattalli, line 10:
```suggestion
def create_person_list(people: list[dict]) -> list[Person]:
```

Nattalli (approved):

PR #1451, opened 2024-10-02
Solution

Rabb1t-Artem:


levseenko (approved):

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list[dict]) -> list:
```
Nattalli, line 10: can you specify type of the returning list items?

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+    person_list = []
+    for person in people_list:
+        person_obj = Person(person["name"], person["age"])
+        if "husband" in person:
+            if person["husband"] is not None:
+                person_obj.husband = person["husband"]
+        if "wife" in person:
+            if person["wife"] is not None:
+                person_obj.wife = person["wife"]
+        person_list.append(person_obj)
```
Nattalli, line 20: can you create person_list with list comprehension? also, logic with creating wife and husband can be implemented in next loop

Nattalli (changes requested):

app/main.py
Rabb1t-Artem: sorry. my bad. my finish code was in other file.

app/main.py
```diff
@@ -1,8 +1,28 @@
...
-    pass
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    person_list = []
+
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+        person = Person(name, age)
+        person_list.append(person)
```
Nattalli, line 17: did you push the last version of your code? you can replace that code with list comprehension that would be more readable and beautiful 💅🏻

Nattalli (changes requested):

PR #1452, opened 2024-10-02
'Solution'

levseenko:


yulia-pl (commented):
I like your code :)

Le0n-K:
good)

Le0n-K (approved):

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+
+def create_person_list(people: list[dict]) -> list["Person"]:
+    person_instances = {}
+
+    for person_data in people:
+        name = person_data["name"]
+        age = person_data["age"]
+
+        person_instance = Person(name, age)
+        person_instances[name] = person_instance
```
Nattalli, line 27: can you use dict comprehension for this?

app/main.py
```diff
@@ -1,8 +1,38 @@
...
+
+    for person_data in people:
+        name = person_data["name"]
+        partner_key = "wife" if "wife" in person_data else "husband"
+        partner_name = person_data.get(partner_key)
+
+        if partner_name:
+            partner_instance = person_instances[partner_name]
+            person_instances[name].set_partner(partner_instance)
```
Nattalli, line 36: can you simplify it? also using additional method is unnecessary because it can be implemented here as you are already doing checking here

Nattalli (changes requested):

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+        partner_key = "wife" if "wife" in person_data else "husband"
+        partner_name = person_data.get(partner_key)
+        if partner_name:
+            partner_instance = person_instances[partner_name]
+            setattr(person_instances[name], partner_key, partner_instance)
+            setattr(
+                partner_instance,
+                "husband" if partner_key == "wife" else "wife",
+                person_instances[name]
+            )
```
Nattalli, line 29: as you iterate over every person in the`person_instances` you can set only attributes for a certain person - `person_data`

Nattalli: also simple `person` or `human` probably would be better then `person_data`

Nattalli (changes requested):

app/main.py
```diff
@@ -15,17 +15,17 @@ def create_person_list(people: list[dict]) -> list["Person"]:
...
+            person_instance = person_instances[person["name"]]
             partner_instance = person_instances[partner_name]
-            setattr(person_instances[name], partner_key, partner_instance)
+            setattr(person_instance, partner_key, partner_instance)
             setattr(
                 partner_instance,
                 "husband" if partner_key == "wife" else "wife",
-                person_instances[name]
+                person_instance
             )
```
Nattalli, line 29: just remove that line, because you set a partner for every person twice, for example:
you have Anna with her husband Ivan, so you are setting Ivan as Anna's husband and Anna as Ivan's wife. And on the next step, you set again: Anna is Ivan's wife and Ivan is Anna's husband

So you better set only Ivan as Anna's husband and in the next step Anna as Ivan's wife

Nattalli (changes requested):

app/main.py
levseenko: Thanks!

Nattalli (approved):
Good job!

PR #1454, opened 2024-10-03
Solution

yulia-pl:


FursykIvan (approved):

app/main.py
```diff
@@ -1,8 +1,31 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
```
Nattalli, line 10: can you specify list items data type?

app/main.py
```diff
@@ -1,8 +1,31 @@
...
-    # write your code here
-    pass
+    Person.people.clear()
```
Nattalli, line 11: why? 😅

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+
+    person_list = []
+
+    for person_dict in people:
+        person = Person(name=person_dict["name"], age=person_dict["age"])
+        person_list.append(person)
```
Nattalli, line 17: use list comprehension here

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+
+    for person_dict in people:
+        person = Person.people[person_dict["name"]]
+        if "wife" in person_dict and person_dict["wife"] is not None:
```
Nattalli, line 21: you can avoid double checking with the usage of the `.get()` method

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+            person.wife = Person.people[person_dict["wife"]]
+        elif hasattr(person, "wife"):
+            delattr(person, "wife")
```
Nattalli, line 24: why?

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+
+        if "husband" in person_dict and person_dict["husband"] is not None:
+            person.husband = Person.people[person_dict["husband"]]
+        elif hasattr(person, "husband"):
+            delattr(person, "husband")
+
```
Nattalli, line 30: the same that on the previous 2 comments

Nattalli (changes requested):

app/main.py
yulia-pl: for data update: If create_person_list is called multiple times with different data, clearing Person.people allows to remove previous data to avoid mixing old and new data.

Arsen-hrynevych (approved):
ok

PR #1484, opened 2024-10-23
Solution

dottfmar:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
```
petrykivd: It's not a good idea to add its type to the variable name; it's enough to just name it in plural.

petrykivd: Change it in other places too

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    for person_dict in people:
+        person = Person(person_dict["name"], person_dict["age"])
+        person_list.append(person)
```
petrykivd: You can also simplify this construction a bit by using list comprehension.

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    for person_dict in people:
+        person = Person.people[person_dict["name"]]
+        if "wife" in person_dict and person_dict["wife"] is not None:
```
petrykivd: You can simplify the check by using the .get() dictionary method. If the dictionary does not have such a key, it will return None. Don't forget to simplify this check elsewhere as well.

petrykivd (changes requested):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = [
+        Person(person_data["name"], person_data["age"])
+        for person_data in people
+    ]
+    for person_data in people:
```
L1nk3rrr: Does `person_data` is the same?  looks like not, because at first situation it's a dict, in a second it instance of Person class.
Make better naming at that situation

app/main.py
```diff
@@ -1,8 +1,21 @@
...
-    pass
+    people = dict()
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
```
L1nk3rrr: You can improve your annotation, because you accept a list of dict and return list of persons.
Try `list[TYPE_NAME]`

L1nk3rrr (changes requested):

petrykivd (approved):
Well done! 🚀

PR #1528, opened 2024-11-08
Solution

Chukhno-Maksym:


app/main.py
```diff
@@ -1,8 +1,29 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    person_list = []
+
+    for peoples in people:
+        person_list.append(Person(peoples["name"], peoples["age"]))
```
petrykivd: You can also simplify this construction a bit by using list comprehension.

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    for peoples in people:
+        person = Person.people[peoples["name"]]
+        spouse_name = peoples.get("wife") or peoples.get("husband")
+        if spouse_name:
+            spouse = Person.people.get(spouse_name)
+            if spouse:
+                if "wife" in peoples:
+                    person.wife = spouse
+                else:
+                    person.husband = spouse
```
petrykivd: This number of internal IFs is not a good solution. Try to simplify this solution, by removing spouse_name and spouse.

app/main.py
petrykivd: It's not a good idea to add its type to the variable name; it's enough to just name it in plural.

petrykivd (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
-    pass
+
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for peoples in people:
+        person = Person.people[peoples["name"]]
+        if "wife" in peoples and peoples["wife"]:
+            person.wife = Person.people.get(peoples["wife"])
+        elif "husband" in peoples and peoples["husband"]:
+            person.husband = Person.people.get(peoples["husband"])
```
petrykivd: You can simplify this check by using the .get() method. Please use checklist before pushing your solution)

petrykivd (changes requested):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+    person_list = [Person(person["name"], person["age"]) for person in people]
+
+    for peoples in people:
```
L1nk3rrr: looks like a bad variable name `peoples` because your dict contains only  data of one person

.flake8
```diff
@@ -4,4 +4,4 @@ ignore = E203, E266, W503, ANN002, ANN003, ANN101, ANN102, ANN401, N807, N818
 max-line-length = 79
 max-complexity = 18
 select = B,C,E,F,W,T4,B9,ANN,Q0,N8,VNE
-exclude = venv, tests
+exclude = venv, tests, .venv
```
L1nk3rrr: Don't push such changes

L1nk3rrr (commented):
Well done! Look at my comments, have a minor changes for u

L1nk3rrr (changes requested):
Well done! Look at my comments, have a minor changes for u

L1nk3rrr (approved):

PR #1535, opened 2024-11-13
solution

julia4406:


app/main.py
```diff
@@ -1,8 +1,30 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        # when obj created add link on it to the dict "people"
```
petrykivd: Please pay attention to the checklist, there is no need to add comments to the code as it is clear and self-explanatory.

app/main.py
```diff
@@ -1,8 +1,30 @@
...
-    pass
+    persons = []
+    for human in people:
+        person = Person(name=human["name"], age=human["age"])
+        persons.append(person)
+        # temporary set attributes "husband" and "wife" if they exist
+        if "wife" in human.keys() and human["wife"] is not None:
+            person.wife = human["wife"]     # set name as string
+        elif "husband" in human.keys() and human["husband"] is not None:
+            person.husband = human["husband"]       # set name as string
```
petrykivd: You can simplify the check by using the .get() dictionary method. If the dictionary does not have such a key, it will return None.

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+
+    for person in persons:
+        # set link on wife(husband) object
+        # for each object "person" who has this attributes
+        if hasattr(person, "wife"):
+            person.wife = person.people.get(person.wife)
+        elif hasattr(person, "husband"):
+            person.husband = person.people.get(person.husband)
```
petrykivd: I think we can combine these cycles into one and simplify the overall code.

petrykivd (changes requested):

app/main.py
julia4406: If you meant to unite both cycles 'for' I can't. Cause first cycle creates all objects,  the second cycle searches instances in list created early. For ex. when 1st object in list created, and it has attribute 'wife' we can't associate this wife-name with object, cause we didn't create it yet and we have to go through  the whole  cycle.
If I'm wrong, please let me know.

README.md
```diff
@@ -1,4 +1,4 @@
-# Class Person
+c# Class Person
```
L1nk3rrr: No need to change README file

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        if human.get("husband"):
+            person.husband = human.get("husband")
+        persons.append(person)
+
+    for person in persons:
+        # set link on wife(husband) object
+        if hasattr(person, "wife"):
+            person.wife = person.people.get(person.wife)
+        elif hasattr(person, "husband"):
+            person.husband = person.people.get(person.husband)
```
L1nk3rrr: You can simplify it for a one `for` loop instead of two. 
Minor tip (as I see Dmytro also left same comment as me) - think about what is going on when you are creating an instance of `Person` and what `people` variable do 🤔.
Try, if any questions, at QnA session we are open for discussion!

L1nk3rrr (changes requested):

app/main.py
L1nk3rrr: Also, `for` loop and list comprehension not the same concept from my point of view

L1nk3rrr (approved):
Nice, you did it 😎
Well done!

PR #1545, opened 2024-11-16
Solution to py-person-class 

Trynadis:
In this pull request, I created:
- class Person with attribute people.
- create_person_list(), this function takes a list of people and returns a list with Person instances instead of dicts.


Alkaponees (approved):
LGMT!

PR #1546, opened 2024-11-16
solution

VovaGld:


app/main.py
```diff
@@ -1,8 +1,18 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result_list = [Person(person["name"], person["age"]) for person in people]
+
+    for num, person in enumerate(people):
+        if "wife" in person and person["wife"] is not None:
+            result_list[num].wife = Person.people[person["wife"]]
+        if "husband" in person and person["husband"] is not None:
```
petrykivd: You can simplify the check by using the .get() dictionary method. If the dictionary does not have such a key, it will return None.

app/main.py
```diff
@@ -1,8 +1,18 @@
...
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result_list = [Person(person["name"], person["age"]) for person in people]
+
+    for num, person in enumerate(people):
```
petrykivd: In this case, you do not need to use an enumerator and access the list by index, it is better to iterate through the list and address a specific object (person).

petrykivd (changes requested):

petrykivd (approved):
Well done! 🚀

PR #1572, opened 2024-11-28
Solution

bodiakof:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        Person(person["name"],
+               person["age"])
+        for person in people
+    ]
+
+    for person_instance in person_instances:
+        person_data = next(
+            (person for person in people if person["name"]
+             == person_instance.name),
+            None)
```
julia4406, line 21: Oh, cool - iterators )
But it looks like too complicated. 
If you have checks in  person_data, may be you can use explicit cycle for it, not for person_instances?

julia4406 (approved):

PR #1577, opened 2024-11-29
cache peope

valerii-kashpur:
- added people cache for Person class
- added spouse reference for each people in cache

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -4,23 +4,24 @@ class Person:
     def __init__(self, name: str, age: int) -> None:
         self.name = name
         self.age = age
-        self.people.update({name: self})
+        Person.people.update({name: self})
 
 
 def create_person_list(people: list) -> list:
-    list_of_persons_instances = [
+    persons_instances = [
         Person(person["name"],
                person["age"])
         for person in people
     ]
 
-    for i, person in enumerate(people):
+    for person in people:
+        instances_cache = Person.people
+        person_instance = instances_cache[person["name"]]
         get_wife = person.get("wife")
         get_husband = person.get("husband")
-        person_instance = list_of_persons_instances[i]
         if get_wife:
-            person_instance.wife = person_instance.people[get_wife]
+            person_instance.wife = instances_cache[get_wife]
         elif get_husband:
-            person_instance.husband = person_instance.people[get_husband]
+            person_instance.husband = instances_cache[get_husband]
 
-    return list_of_persons_instances
+    return persons_instances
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people.update({name: self})
```
julia4406, line 8: people is a class variable, you can't use it with 'self', use Person.people.update({name: self}) instead

app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_persons_instances = [
```
julia4406, line 11: suggestion: make this variable's name shorter (just experience, I had a remark from mentor with the same case)

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+        Person(person["name"],
+               person["age"])
+        for person in people
+    ]
+
+    for i, person in enumerate(people):
+        get_wife = person.get("wife")
+        get_husband = person.get("husband")
+        person_instance = list_of_persons_instances[i]
```
julia4406, line 20: You use indexes from one list(people) to take person_instance from another list. 
In my opinion this is not safe, can we be sure in their accordance? May be it's possible to get person_instance in other way?

app/main.py
```diff
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people.update({name: self})
```
julia4406: people is a class variable, you can't use it with 'self', use Person.people.update({name: self}) instead

app/main.py
julia4406: You use indexes from one list(people) to take person_instance from another list.
In my opinion this is not safe, can we be sure in their accordance? May be it's possible to get person_instance in other way?

app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    list_of_persons_instances = [
+        Person(person["name"],
```
julia4406, line 12: suggestion: make this variable's name shorter (just experience, I had a remark from mentor with the same case)

julia4406 (changes requested):

julia4406 (approved):

PR #1578, opened 2024-11-30
Solution

ruslanyemets:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for human in people:
+        person = Person(human["name"], human["age"])
+        person_list.append(person)
+
+    for index, human in enumerate(people):
```
julia4406: What for you are using indexes? You need just human in people in 'for', without enumerating.

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        if "wife" in human and not human["wife"] is None:
```
julia4406: Possible to simplify this with checking only if human.get['wife']: ... You will have those two checks in one.

app/main.py
```diff
@@ -1,8 +1,25 @@
...
+            Person.people[human["name"]].wife\
+                = Person.people[human["wife"]]
+        elif "husband" in human and not human["husband"] is None:
```
julia4406: The same as with 'wife' above.

julia4406 (changes requested):

julia4406 (approved):

julia4406 (approved):

app/main.py
```diff
@@ -1,8 +1,21 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(human["name"], human["age"]) for human in people]
+
+    for human in people:
+        if human.get("wife"):
+            Person.people[human["name"]].wife\
+                = Person.people[human["wife"]]
```
L1nk3rrr: looks like there is no sense to use `\`, it's a one liner

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        elif human.get("husband"):
+            Person.people[human["name"]].husband\
+                = Person.people[human["husband"]]
```
L1nk3rrr: looks like there is no sense to use `\`, it's a one liner

L1nk3rrr (changes requested):
Well done! Minor styling changes for you

petrykivd (approved):
Well done! 🚀

PR #1591, opened 2024-12-02
Solution

Sichkaridze:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+        Person.people[name] = self
+
+
+def create_person_list(people: list[dict]) -> list:
+    for person in people:
+        Person(person["name"], person["age"])
+    for person in people:
+        current_person = Person.people[person["name"]]
+        if "wife" in person and person["wife"]:
```
ruslanyemets: You can simplify this row with checking only 'if person.get("wife"):' and you will have those two checks in one.

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+            current_person.wife = Person.people.get(person["wife"])
+        if "husband" in person and person["husband"]:
```
ruslanyemets: The same as with 'wife' above.

ruslanyemets (changes requested):

app/main.py
Sichkaridze: Great point, thanks/
I'll change my code

ruslanyemets (approved):

PR #1598, opened 2024-12-03
Solution

yelizaveta-zh:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.wife = None
+        self.husband = None
+        Person.people[name] = self
```
Dogober: In this case I had the same trouble due to you're doing Recursion. Becouse during __init__ you're calling Person again and again.
You need to use self in order to have an access to people

like this: self.people.update({
            name: self
        })
 But you can use another similar way

Dogober (changes requested):
The rest part of your code looks cool

app/main.py
```diff
@@ -1,8 +1,26 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.wife = None
+        self.husband = None
```
Dogober: self.wife = None
self.husband = None

There is no need to write this

yelizaveta-zh:
Update recursion

PR #1599, opened 2024-12-03
'Solution'

Dogober:


yelizaveta-zh (approved):
The code looks correct, I have no questions)

PR #1611, opened 2024-12-03
Solution

shretsker:


TretiakHelen (approved):

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    person_list = []
+    for person_dict in people:
+        person = Person(person_dict["name"], person_dict["age"])
+        person_list.append(person)
+
+    for person in person_list:
+        for person_dict in people:
+            if person_dict["name"] == person.name:
+                if "wife" in person_dict:
+                    if person_dict["wife"] is not None:
```
petrykivd: Nested "for loop" it's not good solution. Try to simplify your solution.

petrykivd (changes requested):
Follow Buddy's comments and fix them all.

app/main.py
```diff
@@ -8,19 +8,46 @@ def __init__(self, name: str, age: int) -> None:
 
 
 def create_person_list(people: list) -> list:
-    pass
-    person_list = []
+    person_list = [
+        Person(person_dict["name"],
+               person_dict["age"])
+        for person_dict in people]
+
```
petrykivd, line 15: We need to improve the code style and naming.
You should not specify what type of variable it is in the variable name.

 That's a good option:
persons = [
        Person(person_data["name"], person_data["age"])
        for persond_data in people
        ]

app/main.py
```diff
@@ -8,19 +8,46 @@ def __init__(self, name: str, age: int) -> None:
...
-                        person.wife = Person.people[person_dict["wife"]]
-                if "husband" in person_dict:
-                    if person_dict["husband"] is not None:
-                        person.husband = Person.people[person_dict["husband"]]
+        person = Person.people[person_dict["name"]]
+
+        if "wife" in person_dict and person_dict["wife"]:
+            person.wife = Person.people.get(person_dict["wife"])
+        if "husband" in person_dict and person_dict["husband"]:
+            person.husband = Person.people.get(person_dict["husband"])
```
petrykivd: Simplify your condition, just use .get() method for it.

app/main.py
```diff
@@ -8,19 +8,46 @@ def __init__(self, name: str, age: int) -> None:
...
+#         for person_dict in people:
+#             if person_dict["name"] == person.name:
+#                 if "wife" in person_dict:
+#                     if person_dict["wife"] is not None:
+#                         person.wife = Person.people[person_dict["wife"]]
+#                 if "husband" in person_dict:
+#                     if person_dict["husband"] is not None:
+#                         person.husband = Person.people[
+#                         person_dict["husband"]]
+#     return person_list
```
petrykivd: Remove comments

petrykivd (changes requested):

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+               person_dict["age"])
+        for person_dict in people]
+
+    for person_dict in people:
+        person = Person.people[person_dict["name"]]
+
+        if person.name != "Joey":
+            # solution to:
+            # "AssertionError: Person with "name" Joey
+            # should not have attribute wife"
```
L1nk3rrr, line 22: Okey, that now a mistake from test side :)

Your code must handle all variations that tests runs. You can add check like `if "wife" in person_dict` and on it `if` check a value if it exist by .get() method for example

app/main.py
L1nk3rrr: Okey, you did it, look for a styling side, now it not using best practices, re-check checklists

L1nk3rrr (changes requested):
If you have any troubles with automated tests, write in the QnA chat and ask mentors, don't try to avoid by `if-else`.
You are almost done! Some minor improvements are needed

PR #1622, opened 2024-12-07
'Solution'

FalconiUA:


app/main.py
```diff
@@ -1,8 +1,34 @@
...
+
+def create_person_list(people: list[dict[str, str | int]]) -> list[Person]:
+    person_list = []
+
+    # Перший цикл: створення екземплярів
+    for person_dict in people:
+        name = person_dict["name"]
+        age = person_dict["age"]
+        person = Person(name, age)
+        person_list.append(person)
```
vsmutok: use list comprehension here

vsmutok (changes requested):

app/main.py
FalconiUA: Thanks for the advice! I'll add it in the next commit.

vsmutok (approved):
Good Job!

PR #1623, opened 2024-12-07
solved

Danylo-Ok:


kuinorymes (changes requested):
Hi, You should run flake8 app and fix some issues
Good luck with it)

PR #1652, opened 2024-12-23
implement Person class and create_person_list method

yarzhuravlov:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    people: dict[str, "Person"] = {}
 
+    def __init__(self, name: str, age: str) -> None:
+        self.name = name
+        self.age = age
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+        self.people[self.name] = self
```
oliasymak24: You should access class attribute through class name not self

oliasymak24 (changes requested):

app/main.py
yarzhuravlov: Applied the changes, thanks for the feedback

oliasymak24 (approved):

PR #1659, opened 2024-12-28
implement function "create_person_list" for py-person-class

ihetay:


app/main.py
```diff
@@ -1,8 +1,29 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        if self not in Person.people:
+            Person.people[self.name] = self
```
Dimosphen1: There is no need to perform this check, assign the value directly

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    for person in people:
+        person_list.append(Person(name=person["name"], age=person["age"]))
+
+    for person in people:
+        current_person = Person.people[person["name"]]
+
+        if "wife" in person and person["wife"]:
+            wife_instance = Person.people.get(person["wife"])
+            if wife_instance:
+                current_person.wife = wife_instance
```
Dimosphen1: You could avoid using nested if condition (because you've created all instances in the previous loop) and write if condition more efficiently, e.g.
```suggestion
        if wife_name := person.get("wife"):
            current_person.wife = Person.people[wife_name]
```

Dimosphen1: Rewrite if condition for husband as well

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+             into a corresponding Person object.
+             If "wife" or "husband" is specified, links to the corresponding
+             Person objects are established.
+    """
+    person_list = []
+    # creating person instances and adding to person_list
+    for person in people:
+        person_list.append(Person(name=person["name"], age=person["age"]))
+    # checking for the presence of the key wife or husband and if found,
+    # assign the value wife or husband to the instance
```
Dimosphen1, line 30: There is no need to add such comments in the future since the code is self-descriptive enough

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #1665, opened 2024-12-30
solution

Baranotik15:


app/main.py
```diff
@@ -1,8 +1,44 @@
...
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+
+        Person.people[name] = self
+
+    def set_wife(self, wife: dict) -> None:
+        """Присваиваем жену"""
```
oliasymak24: This is bad we are using only english in the code

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+
+
+def create_person_list(person_data: list) -> list:
+    """Создает список из всех элементов словаря"""
+    person_list = []
+    for data in person_data:
+        person = Person(data["name"], data["age"])
+        person_list.append(person)
+
+        # Присваиваем супругов
```
oliasymak24: don't leave comments

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+        self.husband = husband
+        husband.wife = self
+
+
+def create_person_list(person_data: list) -> list:
+    """Создает список из всех элементов словаря"""
+    person_list = []
+    for data in person_data:
+        person = Person(data["name"], data["age"])
+        person_list.append(person)
```
oliasymak24: Try to rewrite in one line. Hint: list comprehension

app/main.py
```diff
@@ -1,8 +1,44 @@
...
+
+        # Присваиваем супругов
+        if "wife" in data and data["wife"]:
```
oliasymak24: Use just dict.get() method in the condition, instead of this long one, check everywhere

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+def create_person_list(person_data: list) -> list:
+    """Creates a list of all dictionary elements"""
+    persons = {
+        data["name"]: Person(data["name"], data["age"]) for data in person_data
+    }
+
+    for data in person_data:
+        person = persons[data["name"]]
+
+        if "wife" in data and data["wife"]:
```
oliasymak24: If you are using get() method you don't need `if` before

app/main.py
```diff
@@ -1,8 +1,41 @@
...
+            wife = persons.get(data["wife"])
+            if wife:
+                person.set_wife(wife)
 
+        if "husband" in data and data["husband"]:
```
oliasymak24: Same here

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1676, opened 2025-01-07
Implement create_person_list function to manage relationships

kram3ko:
link Person instances based on input data.


Nikita-pyth (approved):
I like this solution a lot

PR #1717, opened 2025-01-27
Solution

chiper9:


app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        name = person_data["name"]
+        person_instance = Person.people[name]
+
+        if "wife" in person_data and person_data["wife"] is not None:
+            wife_name = person_data["wife"]
+            person_instance.wife = Person.people[wife_name]
+
+        if "husband" in person_data and person_data["husband"] is not None:
+            husband_name = person_data["husband"]
+            person_instance.husband = Person.people[husband_name]
```
MNDonut: you .get method of dict to check if statements

MNDonut (changes requested):

app/main.py
```diff
@@ -17,12 +17,12 @@ def create_person_list(people_data: list) -> list:
         name = person_data["name"]
         person_instance = Person.people[name]
 
-        if "wife" in person_data and person_data["wife"] is not None:
-            wife_name = person_data["wife"]
+        wife_name = person_data.get("wife")
+        if wife_name is not None:
```
Nattalli: also can suggest you something like this:
```python

File: app/main.py [hunk 3]
```diff
@@ -17,12 +17,12 @@ def create_person_list(people_data: list) -> list:
...
             person_instance.wife = Person.people[wife_name]
 
-        if "husband" in person_data and person_data["husband"] is not None:
-            husband_name = person_data["husband"]
+        husband_name = person_data.get("husband")
+        if husband_name is not None:
```
Nattalli: the same here

Nattalli (changes requested):

MNDonut (approved):

PR #1734, opened 2025-02-04
Solution

kbatrakov:


vasylhnatiuk (changes requested):
fix flake8 tests

app/main.py
```diff
@@ -23,11 +21,13 @@ def create_person_list(people: list) -> list:
...
-        for value in Person.people.values():
-            if value.wife:
-                if instance.name == value.wife:
-                    value.wife = instance
-            if value.husband:
-                if instance.name == value.husband:
-                    value.husband = instance
+        if "wife" in instance.__dict__:
+            for partner in instances:
+                if instance.__dict__["wife"] == partner.name:
```
vasylhnatiuk: why did you change it like that?

vasylhnatiuk: use atribute unstead of `__dict__`

vasylhnatiuk (changes requested):

app/main.py
kbatrakov: > why did you change it like that?

I used it because it allowed me to set husband/wife attribute to class instances without setting self.wife/husband = None inside the constructor of the Person class. If i remove __dict__ method and try to use atributes directly it will cause a raising of the Attribute error. If I set self.wife/husband = None inside the constructor, Person class instance "Joey" will have a value None for his "wife" attribute which should not be existing at all according to the conditions of the current task (FAILED tests/test_main.py::test_create_person_list_has_no_wife - AssertionError: Person with 'name' Joey should not have attribute wife). I was just trying to fit in test requirements. Please, correct me if I am wrong.

kbatrakov: @vasylhnatiuk Hopefully, option below is a better one :)

app/main.py
```diff
@@ -1,8 +1,39 @@
...
+            for partner in instances:
+                if instance.wife == partner.name:
+                    instance.wife = partner
+                    break
+                if instance.husband == partner.name:
+                    instance.husband = partner
+                    break
+        else:
+            del instance.wife
+            del instance.husband
```
vsmutok: Your code looks quite complex for this task; please try to simplify it.

vsmutok (changes requested):

SerhiyTo (approved):

PR #1756, opened 2025-02-12
Creating a class and function that creates dependencies

Denisvv1:


app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        self.name = name
+        self.age = age
+        self.__class__.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for i in people:
```
vasylhnatiuk, line 12: i is a bad name, rename pls

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        person = Person(i["name"], i["age"])
+        result.append(person)
+
+    for i in people:
```
vasylhnatiuk, line 16: i is a bad name, rename pls

vasylhnatiuk (changes requested):

PR #1774, opened 2025-02-17
Person_class_Solution

Kart0l:


PrimeGlorious (approved):
Code looks good! Just note that comments in non-English might trigger warnings in pytest. It's better to use English for comments to avoid that.

PR #1776, opened 2025-02-17

PrimeGlorious:


iSevenpwnz (approved):

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+
+    for human in people:
+        if human["name"] not in Person.people:
+            Person(human["name"], human["age"])
+
+    for human in people:
+        person = Person.people[human["name"]]
+
+        if human.get("wife") is not None:
+            person.wife = Person.people.get(human["wife"])
```
oliasymak24: You are doing the same work twice get(human["wife"]
Better to call that function before `if` and check `if wife: // your code`
Check everywhere

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1781, opened 2025-02-18
Solution

LLevkovych:


OlenVe (approved):
Well done!

app/main.py
```diff
@@ -1,8 +1,25 @@
...
-    # write your code here
-    pass
+    persons_list = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+
+        if person.get("wife") is not None:
+
+            Person.people[person.get("name")].wife \
+                = Person.people[person.get("wife")]
```
oliasymak24: Ypu are calling person.get("wife") twice, why not first before if make
`wife = person.get("wife")` then check `if wife: // your code`
Check in all similar situations
Currently your code is bad optimized

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1791, opened 2025-02-19
Solution

DHlomozda:


pewlay:
Good Job! I also think it would be a good idea to make a mutual connection between partners, like 

            Person.people[person["name"]].wife = Person.people[person["wife"]]
            Person.people[person["wife"]].husband = Person.people[person["name"]]

pewlay (approved):

PR #1797, opened 2025-02-20
Solution

ntkachenko21:


yarms1 (approved):

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        Person.people[name] = self
+
+
+def create_person_list(people_data: list) -> list:
+    users_list = []
+    for data in people_data:
+        name = data["name"]
+        age = data["age"]
+        person = Person(name, age)
+        users_list.append(person)
```
oliasymak24: Better to use list comprehension

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+
+    for data in people_data:
+        name = data["name"]
+        person = Person.people[name]
+        if "wife" in data and data["wife"] is not None:
```
oliasymak24: Use get() method here before `if`, then you can avoid duplicated code, check in next `if` too

oliasymak24 (changes requested):

Y-Havryliv (approved):

PR #1801, opened 2025-02-20
solution

Anollinna:


Roksolana-K (approved):
Good job! The code is clean and readable. 

However, I have a few suggestions:  
 - it's better to write "human" instead of "humen".
 - also, as I can see, your code is very similar to mine and AI didnt approved mine on line 22, 25 and 29 __(current_person = Person.people.get(person_data["name"])_)_ with the message: _"There is a potential issue here if person_data["name"] does not exist in Person.people. This could lead to a KeyError. Consider using a default value or checking if the key exists before accessing it."_
 More information about the error here:  https://mate.academy/chats/342230?selectedMessageId=5375127&threadId=5375069

PR #1820, opened 2025-02-27
add task solution

MishaLehotskyi:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = [Person(person["name"], person["age"]) for person in people]
+
+    for person in people:
+        if "wife" in person:
+            if person["wife"] is not None:
+                Person.people[person["name"]].wife\
+                    = Person.people[person["wife"]]
```
vsmutok: better to use `()` then `\`

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        elif "husband" in person:
+            if person["husband"] is not None:
+                Person.people[person["name"]].husband\
+                    = Person.people[person["husband"]]
```
vsmutok: better to use `()` then `\`

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1924, opened 2025-04-24
Solution

Nib20:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
+    for person in people:
+        person_data = Person(person["name"], person["age"])
+        person_instance.append(person_data)
+
+    for person in people:
+        person_data = Person.people[person["name"]]
+        if "wife" in person and person["wife"] is not None:
+            person_data.wife = Person.people[person["wife"]]
+        elif "husband" in person and person["husband"] is not None:
+            person_data.husband = Person.people[person["husband"]]
```
MNDonut: you can use .get() method of dict that makes both of the checks you do in both statements

MNDonut (changes requested):

app/main.py
Nib20: fixed

MNDonut (approved):

PR #1930, opened 2025-04-27
solution

Hetmarum:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+            for person_inst in person_insts:
+                if person_inst.name == person.get("wife", None):
+                    person_obj.wife = person_inst
+                    person_inst.husband = person_obj
+                elif person_inst.name == person.get("husband", None):
+                    person_obj.husband = person_inst
+                    person_inst.wife = person_obj
+        person_insts.append(person_obj)
+    Person.people.update({person_obj: person_inst})
+    return person_insts
```
AnyoneClown: Your alg is to complex. N^2(because of loop inside of another loop).

Refactor without using nested loops

AnyoneClown (changes requested):

MNDonut (approved):

PR #2018, opened 2025-06-17
Solution

DrGregoryHouse4676:


app/main.py
```diff
@@ -1,8 +1,31 @@
...
-    # write your code here
-    pass
+
+def create_person_list(people_list: str) -> list:
+    instance = []
+    Person.people.clear()
+
+    for person_data in people_list:
+        person = Person(name=person_data["name"], age=person_data["age"])
+        instance.append(person)
```
AnyoneClown: Refactor using list comprehension

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: list) -> list:
+    Person.people.clear()
```
MNDonut: what if you get two lists? It means people from the first list will be removed

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    instance = [Person(name=person_data["name"],
+                       age=person_data["age"])
+                for person_data in people_list]
```
MNDonut: move the construction on the next line so that in can fit in one line or
make it in this way:
```
[
    Person(
        name=...,
        age=...
    ) for person in person_list
]
```

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+
+    for person_data in people_list:
+        person = Person.people[person_data["name"]]
+
+        if "wife" in person_data and person_data["wife"] is not None:
+            wife_person = Person.people[person_data["wife"]]
+            person.wife = wife_person
+            wife_person.husband = person
+
+        if "husband" in person_data and person_data["husband"] is not None:
```
MNDonut: use `get` method instead in if statements

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people_list: dict) -> Person:
+    Person.people.clear()
```
MNDonut: my question is still open

app/main.py
```diff
@@ -1,8 +1,36 @@
...
+        if wife_name:
+            wife_person = Person.people[wife_name]
+            person.wife = wife_person
+            wife_person.husband = person
+
+        husband_name = person_data.get("husband")
+        if husband_name:
+            husband_person = Person.people[husband_name]
+            person.husband = husband_person
+            husband_person.wife = person
```
MNDonut: No need to do double assign because you loop through all objects, so at some point you'll get wife or husband of the current person

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,35 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    def __repr__(self) -> None:
+        return f"Person(name={self.name!r}, age={self.age!r})"
+
+
+def create_person_list(people_list: list[dict]) -> list[Person]:
+    Person.people.clear()
```
AnyoneClown: Check line 14 and previous comments from Ihor

AnyoneClown (changes requested):

AnyoneClown (changes requested):
You didn't fix previous comment

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        Person(person["name"], person["age"])
```
L1nk3rrr: here you can do a list comprehension and save it to some variable such a list_of_persons or result and simply return it, cause now you did two actions instead of 1.

```
result = [Person(person["name"], person["age"]) for person in people]
....
....
your_code
....
....
return result
```

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+
+    for person in people:
+        current_person = Person.people[person["name"]]
+
+        if "wife" in person and person["wife"] is not None:
```
L1nk3rrr: also you ignored comment about using `get`

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+            wife_name = person["wife"]
+            if wife_name in Person.people:
+                current_person.wife = Person.people[wife_name]
+
+        if "husband" in person and person["husband"] is not None:
```
L1nk3rrr: same here

L1nk3rrr (changes requested):
Please review comments, also not forget about previous ones

AnyoneClown (approved):
LGTM

PR #2022, opened 2025-06-19
my_solution

Orixara:


zqpiq (approved):

Orixara:
@mateacademy-ai-mentor 
I have reviewed your comments, and I'd like to clarify a few points that were misunderstood:
element person can't have 2 attributes "wife" and "husband" at the same time, because one person can have or wife or husband or None, but two in a row not, so can you re-review my task?

PR #2026, opened 2025-06-19
Solution

Gleg0:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
-    pass
+    for person_in_list in people:
+        Person(person_in_list["name"], person_in_list["age"])
+
+    for person_in_list in people:
+        person = Person.people[person_in_list["name"]]
+        if "wife" in person_in_list and person_in_list["wife"]:
+            person.wife = Person.people[person_in_list["wife"]]
+        if "husband" in person_in_list and person_in_list["husband"]:
+            person.husband = Person.people[person_in_list["husband"]]
```
yagginov: instead of checking the type:
`"wife" in person`

we are advised to use
`person.get("wife")`

app/main.py
```diff
@@ -1,8 +1,21 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person_in_list in people:
```
yagginov: Regarding naming, I think it would be clearer to just say `person`

yagginov (changes requested):

yagginov (approved):

app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
```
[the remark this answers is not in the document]
yagginov: where the condition states that Person.people should be reset with each call to the create_person_list function?

PR #2027, opened 2025-06-20
Solution for Person class

yagginov:


app/main.py
```diff
@@ -1,8 +1,51 @@
...
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
```
DrGregoryHouse4676, line 8: Not needed after writing real code(PASS)

app/main.py
```diff
@@ -1,8 +1,51 @@
...
+    people = {}
 
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
```
DrGregoryHouse4676, line 6: def create_person_list(people: list) You are calling the function twice, why?

DrGregoryHouse4676 (changes requested):
Deal with def create_person_list(people: list) -> list.... I can't understand what function it is for "cancel_marriage" and "get_married"

DrGregoryHouse4676 (approved):
Good job!

app/main.py
yagginov: As I understand it, you wrote comments not to my code, but to the original one)))

PR #2029, opened 2025-06-20
Solution

yuliiakutsa:


yuliiakutsa (commented):
Added dictionary clearing before creating a new dic

yuliiakutsa (commented):
Added func annotation

PR #2040, opened 2025-06-22
implemented class Person and create_person_list func

dimashevchukk:


SerhiiBasok (approved):
Tests successful, no errors found

PR #2041, opened 2025-06-22
create func create_person_list

SerhiiBasok:


DrGregoryHouse4676 (approved):
Don't forget to add readability to the code, because everything is crammed together and there is not a single free line.

PR #2054, opened 2025-06-24
Solution

MakovM:


SensibleN00B (changes requested):
There is one critical issue that needs to be addressed before I can approve your PR: the class-level Person.people dictionary is not cleared at the start of create_person_list. This means that if the function is called multiple times, old data will persist and could cause incorrect associations between people. To fix this, please clear the Person.people dictionary at the beginning of the function.

PR #2059, opened 2025-06-24
solution

MorsImmortalis05:


Bebaaaaa (approved):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for person in people:
+        name = person["name"]
+        age = person["age"]
+        result.append(Person(name, age))
+
+    for index, person in enumerate(people):
```
L1nk3rrr: why we use enumerate here?

app/main.py
```diff
@@ -1,8 +1,23 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
```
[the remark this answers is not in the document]
L1nk3rrr: totaly agree, at start of `create_person_list` please clear a people dict

app/main.py
```diff
@@ -1,8 +1,23 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for person in people:
+        name = person["name"]
+        age = person["age"]
+        result.append(Person(name, age))
```
L1nk3rrr: it looks like here can be used a list comperehension

app/main.py
```diff
@@ -1,8 +1,23 @@
...
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int, **kwargs) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
```
L1nk3rrr, line 10: we accept list of dicts (`list[dict]`) and return list of Persons (`list[Person]`) 

Also to use a class in annotation import `from __future__ import annotations`

L1nk3rrr (changes requested):
Well done! Minor comments for improvements left

app/main.py
MorsImmortalis05: I tried to solve some issue using index and forgot to remove enumerate(

AnyoneClown (approved):
LGTM

PR #2068, opened 2025-06-25
created class Person

trellle:


app/main.py
```diff
@@ -1,8 +1,30 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people.update({self.name: self})
```
MNDonut: you can add a value to the dict just via [] brackets

app/main.py
```diff
@@ -1,8 +1,30 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for index in range(0, len(people)):
+        result.append(Person(people[index]["name"], people[index]["age"]))
```
MNDonut: why do you use a raw loop if you can use either
1. list comprehension
2. loop through objects, not indexes?

app/main.py
```diff
@@ -1,8 +1,30 @@
...
+                    result[num].husband = result[index]
+                    result[index].wife = result[num]
+        elif ("husband" in people[index]
+              and people[index]["husband"] is not None
+              and not hasattr(result[index], "husband")):
+            result[index].husband = people[index]["husband"]
+            for num in range(0, len(result)):
+                if result[num].name is result[index].husband:
+                    result[num].wife = result[index]
+                    result[index].husband = result[num]
```
MNDonut: It's a c++ style, we don't write code like this in Python
1. the same thing, you don't need to loop through indexes because in python you can loop through objects
2. This condition
```
("husband" in people[index]
and people[index]["husband"] is not None
and not hasattr(result[index], "husband"))
```
can be replaced by `.get()` method

3. Even if you use indexes, then make them more descriptive because it's difficult to understand what num and index mean here and why you add nested loops
4. You don't need nested loops. Just assign a wife/husband to the current object. Your init adds all objects in people dictionary, so you can just fetch any wife/husband from there

MNDonut (changes requested):

app/main.py
trellle: I didn't know about .get() method, I'll check it out

trellle: >Just assign a wife/husband to the current object. Your init adds all objects in people dictionary, so you can just fetch any wife/husband from there

I made that loop in order to prevent double-assigning, when I assign a husband for woman, I search for him and assisn him a wife.

trellle: I also use index because it's used in people[index] and result[index] in order to combine them in one loop

trellle: if I use "for elem in result" construction, can I change result's elements though elem?

MNDonut: 1. You can use if statements to check if you need to assign wife or husband
2. No need to use indexes at least in this task
3. Yes, using `elem.wife = ...`

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+                    elem.husband = result[index]
+                    result[index].wife = elem
+        elif ("husband" in people[index]
+              and people[index]["husband"] is not None
+              and not hasattr(result[index], "husband")):
+            result[index].husband = people[index]["husband"]
+            for elem in result:
+                if elem.name is result[index].husband:
+                    elem.wife = result[index]
+                    result[index].husband = elem
```
MNDonut: 1. Why do you use index?
You can just use person.wife = Person.people[wife_name]
2. Use .get() method.
3. No nested loops required in this task

MNDonut (changes requested):

app/main.py
trellle: I thought I can add values to dict both through the key and through the update() method

app/main.py
trellle: the trouble is that I don't know has a person wife or husband, so I check the corresponding key in people list

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for elem in people:
+        person = Person(elem["name"], elem["age"])
```
AnyoneClown: Use better naming for all your variables

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+        result.append(person)
+        person = None
+    for elem1 in result:
+        for elem2 in result:
+            if hasattr(elem1, "wife") and elem1.wife is elem2.name:
+                elem2.husband = elem1
+                elem1.wife = elem2
+            elif hasattr(elem1, "husband") and elem1.husband is elem2.name:
+                elem2.wife = elem1
+                elem1.husband = elem2
```
AnyoneClown: You have a loop inside of another loop. Refactor only using 1

app/main.py
```diff
@@ -1,8 +1,28 @@
...
-    pass
+    result = []
+    for elem in people:
+        person = Person(elem["name"], elem["age"])
+        if elem.get("wife"):
+            person.wife = elem["wife"]
+        elif elem.get("husband"):
+            person.husband = elem["husband"]
+        result.append(person)
+        person = None
```
AnyoneClown: Use list comprehension here

AnyoneClown: You don't have to set `person` to `None`

app/main.py
```diff
@@ -1,8 +1,28 @@
...
+    for elem1 in result:
+        for elem2 in result:
+            if hasattr(elem1, "wife") and elem1.wife is elem2.name:
+                elem2.husband = elem1
+                elem1.wife = elem2
+            elif hasattr(elem1, "husband") and elem1.husband is elem2.name:
```
AnyoneClown: Refactor using `.get()` method

AnyoneClown (changes requested):

app/main.py
trellle: idk how

app/main.py
trellle: there's no list to use .get()

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+def create_person_list(people_data: list) -> list:
+    people = [Person(person["name"], person["age"]) for person in people_data]
+    for index in range(0, len(people_data)):
+        wife = people_data[index].get("wife")
+        if wife:
+            people[index].wife = Person.people[wife]
+        husband = people_data[index].get("husband")
+        if husband:
+            people[index].husband = Person.people[husband]
+    return people
```
MNDonut: Why do you use indexes again?
You can loop through objects

MNDonut (changes requested):

app/main.py
trellle: no, I can't. Denys said to avoid nested loops

L1nk3rrr (approved):
Well done!

PR #2069, opened 2025-06-25
Solution

reinvvay:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
-    # write your code here
-    pass
+
+def create_person_list(people: list) -> list[Person]:
+    persons = []
+    for person_dict in people:
+        name = person_dict["name"]
+        age = person_dict["age"]
+        person = Person(name, age)
+        persons.append(person)
```
AnyoneClown: Refactor using list comprehension

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,8 +1,20 @@
...
+
+def create_person_list(people: list) -> list[Person]:
+    persons = [Person(p_dict["name"], p_dict["age"]) for p_dict in people]
+
+    for person_dict in people:
+        person = Person.people[person_dict["name"]]
+
+        if "wife" in person_dict and person_dict["wife"] is not None:
+            person.wife = Person.people[person_dict["wife"]]
+        elif "husband" in person_dict and person_dict["husband"] is not None:
```
MNDonut: you can use .get method instead of two checks

MNDonut (changes requested):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
-    pass
+
+def create_person_list(people: list) -> list[Person]:
+    persons = [Person(p_dict["name"], p_dict["age"]) for p_dict in people]
+
+    for person_dict in people:
+        person = Person.people[person_dict["name"]]
+
+        partner_husband = person_dict.get("husband", None)
+        partner_wife = person_dict.get("wife", None)
```
MNDonut: No need to set None if keys is not present. None - is a default value for this method

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+
+        if partner_husband is not None:
+            person.husband = Person.people[partner_husband]
+        if partner_wife is not None:
```
MNDonut: just `if partner`

MNDonut (changes requested):

AnyoneClown (approved):
LGTM

PR #2078, opened 2025-06-30
Solution

a1apek:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+    for entry in data:
+        person = Person(entry["name"], entry["age"])
+        person_list.append(person)
+
+    for entry in data:
+        person = Person.people[entry["name"]]
+        if "wife" in entry and entry["wife"]:
+            setattr(person, "wife", Person.people[entry["wife"]])
+        if "husband" in entry and entry["husband"]:
+            setattr(person, "husband", Person.people[entry["husband"]])
```
MNDonut, line 21: 1. Use .get() method in if statements
2. Instead of setattr use direct assignment via `.wife/husband = ...`

MNDonut (changes requested):

PR #2083, opened 2025-07-02
feat: solution

ArtemZakhar:
Hi)
Have a nice day)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,19 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_persons = [Person(person["name"], person["age"]) for person in people]
+
+    for index, person in enumerate(people):
+        if "wife" in person and person["wife"]:
+            new_persons[index].wife = Person.people[person["wife"]]
+        if "husband" in person and person["husband"]:
+            new_persons[index].husband = Person.people[person["husband"]]
+
+    return new_persons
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,8 +1,19 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    new_persons = [Person(person["name"], person["age"]) for person in people]
+
+    for index, person in enumerate(people):
+        if "wife" in person and person["wife"]:
+            new_persons[index].wife = Person.people[person["wife"]]
+        if "husband" in person and person["husband"]:
+            new_persons[index].husband = Person.people[person["husband"]]
```
MNDonut: 1. You don't need to use indexes here. You can just loop through objects of `new_persons` list and access or change its `wife`/`husband` directly
2. You can simplify if statements using `.get()` method of a dict

MNDonut (changes requested):
Have a nice day too!)

app/main.py
ArtemZakhar: I don't know how to get rid of indexes, but the "get" approach I liked)

MNDonut: Just loop through `people` list objects and fetch objects from `Person.people` dict

app/main.py
```diff
@@ -1,8 +1,20 @@
...
-    # write your code here
-    pass
+    new_persons = [Person(person["name"], person["age"]) for person in people]
+
+    for index, new_person in enumerate(new_persons):
+        person_data = people[index]
+        if person_data.get("wife"):
+            new_person.wife = Person.people[person_data["wife"]]
+        if person_data.get("husband"):
+            new_person.husband = Person.people[person_data["husband"]]
```
MNDonut: here you can loop through `people` objects list directly and get all objects you need from `Person.people` dict instead of using indexes

MNDonut (changes requested):

app/main.py
ArtemZakhar: at last I get it)

oliasymak24 (approved):

PR #2151, opened 2025-08-18
Solution

2d1Corp:


2d1Corp:
" The assignment explicitly requires using a list comprehension to create the list of Person " 
Why its mandatory to use list comperhension if there was no such a requirement in readme? @mateacademy-ai-mentor

PR #2192, opened 2025-09-08
Solution

Vorongor:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
-    pass
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
+
+
+def create_person_list(people: list[Person]) -> list:
+    result = []
+    for human in people:
```
DonDmytron: What do you think about list comprehension in this case?

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        person = Person(human["name"], human["age"])
+        result.append(person)
+
+    for human in people:
+        person = Person.people[human["name"]]
+        if "wife" in human and human["wife"] is not None:
```
DonDmytron: Using the get method will reduce the number of check's

app/main.py
```diff
@@ -1,8 +1,24 @@
...
+            person.wife = Person.people[human["wife"]]
+        if "husband" in human and human["husband"] is not None:
```
DonDmytron: same ↑

DonDmytron (commented):
Please, fix some issues.

DonDmytron (changes requested):
Please, fix some issues.

DonDmytron (approved):
good)

PR #2232, opened 2025-09-29
py-person-class Solution

nklsnykk:


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = []
+    for entry in people:
+        person = Person(entry["name"], entry["age"])
+        persons.append(person)
```
vsmutok: This violates checklist item 3: "When creating a list of instances from a collection (such as a list of dictionaries), it is recommended to use list comprehension rather than traditional `for` loops." The code uses a traditional loop to build `persons` (lines 11-14). Prefer a list comprehension to construct the Person instances so the code follows the checklist guidance.

vsmutok (changes requested):

tests/test_main.py
vsmutok, line 1: please do not change the tests, restore them to their original state

vsmutok (changes requested):

PR #2281, opened 2025-10-21
add solution

VadymVynokur:


app/main.py
```diff
@@ -1,8 +1,23 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+    for person_dic in people:
+        person = Person(person_dic["name"], person_dic["age"])
+        result.append(person)
```
Vorongor: There is a better option to create a list -> list comprehension

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+
+    for person_dic in people:
+        name = person_dic["name"]
+        if "wife" in person_dic and person_dic["wife"]:
+            Person.people[name].wife = Person.people[person_dic["wife"]]
+        if "husband" in person_dic and person_dic["husband"]:
+            Person.people[name].husband = Person.people[person_dic["husband"]]
```
Vorongor, line 21: A better option would be to use the get method for dictionaries, you can avoid double checks

Vorongor (approved):
Overall, the code fulfills its main purpose, but it can be improved in terms of readability and Python best practices. Specifically:

Replace the explicit loop used to create result with a list comprehension for more concise and efficient code.

Use the dict.get() method instead of multiple "key" in dict checks to simplify conditional logic and improve clarity.

These changes will make the code cleaner, more Pythonic, and easier to maintain.

PR #2287, opened 2025-10-24
Added solution

vadym-zahrebelnyi:
Solution without documentation because of strange behavior of some tests


app/main.py
```diff
@@ -1,8 +1,24 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    persons = [
```
leemonch1k-space, line 11: Here if you already have person from list in people{} it will be updated. In this case will be more correct if you check if person already exist in people at first, then initializatin method will be used lesser.

leemonch1k-space (approved):
Task done well. Great job!

PR #2288, opened 2025-10-25
Solution

leemonch1k-space:


app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    for person in people:
+        if person.get("name") not in Person.people:
+            Person(person.get("name"), person.get("age"))
```
VadymVynokur: it is not necessary to use that if, also better option - create a list -> list comprehension

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+
+    for person in people:
+        method_person = Person.people[person["name"]]
+        if person.get("husband") is not None:
```
VadymVynokur, line 21: if person.get("husband") -> return true by defolt, if not None

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+            method_person.husband = Person.people[person["husband"]]
+        if person.get("wife") is not None:
```
VadymVynokur, line 23: same here, u can just delete "is not None"

VadymVynokur (approved):
Overall, the code fulfills its main purpose, but it can be improved in terms of readability and Python best practices.

app/main.py
leemonch1k-space: done

PR #2293, opened 2025-10-25
Completed py-person-class task

vakt159:


app/main.py
```diff
@@ -1,8 +1,21 @@
 class Person:
-    # write your code here
-    pass
+
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        self.people[self.name] = self
```
Iventyk: Maybe here is better to use Person.people than self.people

Iventyk (approved):
Nice job!

PR #2296, opened 2025-10-26
created people and function create_person_list

AngryAndrii:


app/main.py
```diff
@@ -1,8 +1,26 @@
 class Person:
```
stanislav-kishechko, line 1: I recommend adding documentation to the code.

app/main.py
```diff
@@ -1,8 +1,26 @@
...
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
```
stanislav-kishechko, line 7: In this line, we can refer to the field through an instance of the class using self.

app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    Person.people.clear()
```
stanislav-kishechko, line 11: This line does not make sense because no instance of the People class has been created and it is not a reference to a specific variable.

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+
+    for one_people in people:
+        Person(one_people["name"], one_people["age"])
```
stanislav-kishechko: We can implement this code using list comprehension, as it is not complex logic and it will not affect the readability and understanding of the code.

stanislav-kishechko (approved):

PR #2300, opened 2025-10-27
Solution

Ann-Riabykina:


app/main.py
```diff
@@ -1,8 +1,21 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
```
vadym-zahrebelnyi, line 12: You could use list comprehension for more pythonic solution, but that works fine too.

vadym-zahrebelnyi (approved):

PR #2306, opened 2025-10-27
Create Person class

n-korobko:


Najttsu (approved):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list["Person"]:
+    Person.people.clear()
+    [Person(data["name"], data["age"]) for data in people]
+
+    for data in people:
+        person = Person.people[data["name"]]
```
petrykivd, line 15: I prefer call person instead of data here, but it's not super important 😃

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+
+        if data.get("wife") is not None:
+            person.wife = Person.people[data["wife"]]
+
+        if data.get("husband") is not None:
+            person.husband = Person.people[data["husband"]]
+
+    return list(Person.people.values())
```
petrykivd, line 23: When you create a list comprehension, you can save it into variable and return here, instead of calling new operation for creating list

petrykivd (changes requested):

PR #2315, opened 2025-10-29
Person class and create_person_list implimentation

AlexMishKo:


ohorodnykostap (approved):
looks good

app/main.py
```diff
@@ -1,8 +1,29 @@
...
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list) -> list[Person]:
+    if not people:
+        return []
+    for person_data in people:
+        Person(name=person_data["name"], age=person_data["age"])
```
DonDmytron, line 13: Use List Comprehension here

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+    person_instances = []
+    for person_data in people:
+        name = person_data["name"]
+        current_instance = Person.people[name]
+        partner_key = "wife" if "wife" in person_data else \
```
DonDmytron: Think about method get. How you can simplify your logic using it.

DonDmytron (changes requested):

app/main.py
AlexMishKo: Used List Comprehension here. need help. made a commit, push but i cant see any changes here on git hub

app/main.py
AlexMishKo: method get implemented

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+
+def create_person_list(people: list) -> list[Person]:
+
+    if not people:
+        return []
+
+    [
+        Person(name=person_data["name"], age=person_data["age"])
+        for person_data in people
+    ]
```
oliasymak24: You' re creating here list comprehension but not set it to any variable and not using it

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+
+    for person_data in people:
+        name = person_data["name"]
+        current_instance = Person.people[name]
+        partner_name = person_data.get("wife") or person_data.get("husband")
+        if partner_name is not None:
```
oliasymak24: `Is not None` is redundant here, since in your case if statement will be True if key exists

app/main.py
```diff
@@ -1,8 +1,32 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[self.name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list) -> list[Person]:
```
oliasymak24, line 10: Try to simplify ur function

oliasymak24 (changes requested):

DonDmytron (approved):
I approve it, but remove lookup from 12 lines and create person_instances, where create Person with name and age instead of 25 line

PR #2362, opened 2025-11-21
Solution

meh-666:


meh-666:
Regarding your comment — “First, the Person.people class dictionary is not cleared between calls to create_person_list.”
My function should not modify the class-level attribute, because it may contain data that did not come from this function and that we are not supposed to touch. Those entries may be used by other functions or methods, so clearing this dictionary could break unrelated logic.

meh-666:
I believe you are not correct about clearing the class attribute.
If we clear the Person.people dictionary inside this function, it may negatively affect other functions or methods that rely on this class-level data. My understanding is that this function should not interfere with the behavior of other class logic — SRP and all that.

Regarding the documentation string, it follows good practices, and there are no issues there, exactly as stated in the checklist. @mateacademy-ai-mentor

vsmutok (approved):
Great!

PR #2406, opened 2025-12-06
'Solution'

Nahorni:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        self.age = age
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+    for person in people:
+        person_list.append(Person(person["name"], person["age"]))
```
vsmutok: you can use list comprehension here

app/main.py
```diff
@@ -1,8 +1,27 @@
...
+    for person in people:
+        name = person["name"]
+        current = Person.people[name]
+        if "wife" in person and person["wife"] is not None:
+            if person["wife"] is not None:
+                wife = Person.people[person["wife"]]
+                current.wife = wife
+                wife.husband = current
+        if "husband" in person and person["husband"] is not None:
+            if person["husband"] is not None:
```
vsmutok: You are doing this check twice here:
person["husband"] is not None
One time shod be enough.

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #2431, opened 2025-12-12
Solution

ostapred:


ostapred (commented):
add Person.people. clear in create_person_list and changed loop for to list competition.

PR #2499, opened 2026-01-13
Class Person and function created_person_list created

titoiriepal:
Tested with pytest


titoiriepal (commented):
I changed  the code for use 'person.get'

PR #2546, opened 2026-01-28
fix: use double quotes and remove russian comments

NiMiR13:


ratatuiii (approved):
Interesting, seems good to go

PR #2593, opened 2026-02-15
Solution py-person-class by Kryvutsa

pawunder1992:


app/main.py
```diff
@@ -1,8 +1,19 @@
...
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(data["name"], data["age"]) for data in people]
+    for index in range(len(people)):
+        if (people[index].get("wife") or people[index].get("husband")
+                in Person.people):
+            person_list[index].wife \
+                = Person.people.get(people[index].get("wife"))
```
AnyoneClown: You can refactor code style for this lines. It's not the best practice to use `\`, even if PEP8 allows it

Example:
```suggestion
            person_list[index].wife = Person.people.get(
                people[index].get("wife")
            )
```

Refactor in all cases below too

app/main.py
```diff
@@ -1,8 +1,19 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[f"{self.name}"] = self
```
AnyoneClown: If we already expect name to be `str`, why do we need to use f-string here?

`Person.people[name] = self` -> better way here

app/main.py
```diff
@@ -1,8 +1,19 @@
...
 
 
 def create_person_list(people: list) -> list:
```
AnyoneClown: In type hint for people we can specify what type of objects should be returned in our `list` In this case It should be `Person` class.

Add it here

app/main.py
```diff
@@ -1,8 +1,19 @@
...
-    # write your code here
-    pass
+    person_list = [Person(data["name"], data["age"]) for data in people]
+    for index in range(len(people)):
+        if (people[index].get("wife") or people[index].get("husband")
+                in Person.people):
```
AnyoneClown: No reason to use `for in range` loop. Use default loop `for person in people`, refactor ur code

AnyoneClown (changes requested):

app/main.py
elija-xp: раджу очіщати person.people після виклику метода

    Person.people.clear()

app/main.py
```diff
@@ -1,8 +1,19 @@
...
+        self.age = age
+        Person.people[f"{self.name}"] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = [Person(data["name"], data["age"]) for data in people]
+    for index in range(len(people)):
+        if (people[index].get("wife") or people[index].get("husband")
```
elija-xp: if (people[index].get("wife")) or (people[index].get("husband") in Person.people)

а у вашомуу випадк умова True, якщо просто існує ключ "wife" — навіть якщо такої людини немає в Person.people. Крім того, завжди встановлюються обидва атрибути (wife і husband), навіть якщо значення None.

elija-xp (changes requested):

app/main.py
pawunder1992: i maked it shorter

app/main.py
pawunder1992: done

app/main.py
pawunder1992: changed

pawunder1992: done

elija-xp (approved):

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    Person.people.clear()
+    [Person(data.get("name"), data.get("name")) for data in people]
```
AnyoneClown: You are creating a person with 2 "name" attributes. It should be `name` and `age` instead.

Don't use List Comprehension, if you are not using list after that. Instead - use `for` loop to iterate through your data

app/main.py
```diff
@@ -1,8 +1,22 @@
...
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
-def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+
+def create_person_list(people: list[dict]) -> list[Person]:
+    Person.people.clear()
```
AnyoneClown: I though comment from a student. But your function is called `create_person`. So you shouldn't remove any peope. Remove this line

AnyoneClown (changes requested):
Overall - good. Only small changes

app/main.py
pawunder1992: done

app/main.py
pawunder1992: yes, i listen a student and wrote that part of code

AnyoneClown (approved):
LGTM, great job

PR #2603, opened 2026-02-17
'Solution'

AlinaSv88:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_ls = []
+
+    for person in people:
+        Person(person["name"], person["age"])
+        person_ls.append(Person.people[person["name"]])
```
vsmutok: You can use a list comprehension here instead of a regular loop

vsmutok (changes requested):

app/main.py
AlinaSv88: Done

DaTrEvTeR (approved):
Well done!

PR #2681, opened 2026-03-19
final_result

antar1x:


app/main.py
```diff
@@ -1,8 +1,29 @@
 class Person:
-    # write your code here
-    pass
+    people = {}
+
+    def __init__(self, name, age):
```
PavelPereverzev1: Missing type annotation for function arguments "name", "age".
Missing return type annotation

app/main.py
```diff
@@ -1,8 +1,29 @@
...
+        Person.people[name] = self
+
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+
+    for p in people:
```
PavelPereverzev1: Using a single letter for a variable name reduces code readability.

PavelPereverzev1 (changes requested):

PR #2710, opened 2026-03-29
solution

Prus-Oleksandr:


alisahoys (approved):
the solution correctly creates Person instances and links wife/husband using the shared Person.people dictionary.

PR #2738, opened 2026-04-01
Solution

BuchArtur:


AnyoneClown (changes requested):
Rejected, made with an AI

app/main.py
```diff
@@ -7,23 +7,15 @@ def __init__(self, name: str, age: int) -> None:
...
-
-    for p_dict in people_data:
-        Person(p_dict["name"], p_dict["age"])
-
-    result = []
-    for p_dict in people_data:
-        name = p_dict["name"]
-        instance = Person.people[name]
-
-        spouse_key = "wife" if "wife" in p_dict else "husband"
```
AnyoneClown: Good, but let's refactor it without using `range` loop. Use simple `for` loop instead

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM

PR #2752, opened 2026-04-06
Solution

General-Kaolin:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
+
+        wife_name = pers.get("wife")
+        if wife_name:
+            person_instance.wife = Person.people[wife_name]
+
+        husband_name = pers.get("husband")
+        if husband_name:
+            person_instance.husband = Person.people[husband_name]
+
+    return [Person.people[pers["name"]] for pers in people]
```
vsmutok: You are using a loop here three times.
Try to do it only twice so it’s more efficient.

vsmutok (changes requested):

PR #2776, opened 2026-04-15
Solution

RWAdelP:


RWAdelP:
@mateacademy-ai-mentor 
The reviews are inconsistent
It's ok to use p in a for loop but not i_person.

PR #2778, opened 2026-04-16
Solution

LeonBecken:


LeonBecken (commented):
fixed sintaxis and listcomprehension

PR #2815, opened 2026-05-06
Сделано задание

Mirash1l1:


app/main.py
```diff
@@ -1,8 +1,25 @@
...
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    result = []
+
+    for dict_people in people:
+        result.append(Person(dict_people["name"], dict_people["age"]))
```
vsmutok, line 14: Use list comprehension instead of a traditional `for` loop with `.append()`. Example: `result = [Person(person["name"], person["age"]) for person in people]`

vsmutok (changes requested):

PR #2826, opened 2026-05-14
work is done

Sardor-Tozhiyev:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person in people:
+        new_person = Person(person["name"], person["age"])
+        person_list.append(new_person)
```
vsmutok: Violates checklist item №3: Use list comprehension instead of traditional for loop when creating a list of instances. Replace with: `person_list = [Person(person["name"], person["age"]) for person in people]`

vsmutok (changes requested):

app/main.py
```diff
@@ -1,8 +1,28 @@
...
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for person in people:
+        person_list = [
+            Person(person["name"],
+                   person["age"]) for person in people
+        ]
```
vsmutok: The `for` loop is unnecessary.

You are recreating the entire list on every iteration and overwriting `person_list` each time. The loop does nothing useful.

You only need the list comprehension:

```python
person_list = [
    for person in people
]
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+    def __init__(self, name: str, age: int) -> None:
+        self.name = name
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
```
vsmutok: Remove this `person_list` variable, as it is unnecessary here.

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #2852, opened 2026-05-27
implemented the Person class and the create_person_list function

kv-young:


app/main.py
```diff
@@ -1,8 +1,27 @@
...
+        self.age = age
+        Person.people[name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    [Person(name=person["name"], age=person["age"]) for person in people]
+
+    result_list = []
```
vsmutok: Try storing the created `Person` objects directly in `result_list` instead of using a list comprehension only for side effects. This way, you create the objects and keep the list in one step.

vsmutok (changes requested):

kv-young:
made the proposed changes

vsmutok (approved):
Good Job!

PR #2863, opened 2026-05-30
Solution is added

RinBd:


RinBd:
According to the AI review (checklist item #1) variable names are changed.

PR #2912, opened 2026-06-19
feat: created func(create_person_list) and class Person

neeeksa:


app/main.py
```diff
@@ -1,8 +1,31 @@
...
-    # write your code here
-    pass
+    person_list = []
+
+    for i in people:
+        name = i["name"]
+
+        person = Person(name, i["age"])
+        Person.people[name] = person
+        person_list.append(person)
```
vsmutok, line 18: use list comprehention here

app/main.py
```diff
@@ -1,8 +1,31 @@
...
+
+    for i in people:
```
vsmutok, line 20: `i` bad variable name

vsmutok (changes requested):

PR #3015, opened 2026-08-07
Solution

Sango-dev:


Sango-dev (commented):
fixed dict.get and added list comprehension

Sango-dev (commented):
fixed dict.get and added list comprehension

PR #3016, opened 2026-08-07
Solution

ArGus20:


app/main.py
```diff
@@ -1,8 +1,26 @@
...
+        Person.people[self.name] = self
 
 
 def create_person_list(people: list) -> list:
-    # write your code here
-    pass
+    person_list = []
+
+    for human in people:
+        person = Person(human["name"], human["age"])
```
Avalanche-one: README file recomends us to use list comprehension in this case

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+        person_list.append(person)
+
+    for human in people:
+        person = Person.people[(human["name"])]
+
+        if "wife" in human and human["wife"] is not None:
```
Avalanche-one: There's no need to use such condition. You can just write "if human["wife"], because if if human["wife"] == None, you will automatically receive "False". Besides, README file recomends us to use get() method in this case

app/main.py
```diff
@@ -1,8 +1,26 @@
...
+            person.wife = Person.people[human["wife"]]
+
+        elif "husband" in human and human["husband"] is not None:
```
Avalanche-one: Same here

Avalanche-one (approved):
Looks good, but it would be even better with README recomendations implemented:)

PR #3025, opened 2026-08-09
Solution

MaksimDonchyk:


ArGus20 (approved):
It works, but you’ve overcomplicated it. You can add Person(name, age) using a list comprehension, and you can simply add the wife/husband with if/elif human.get("wife"/"husband"): ...

PR #3028, opened 2026-08-10
Solution

garamond1281:


CemXokenc (approved):
Everything is fine. Your solution is better than mine =)

PR #3041, opened 2026-08-11
Solution

kateryna-che:


lizagulchenko-collab (approved):
17: if person_data.get("wife") is not None
20: if person_data.get("husband") is not None
You don't really need "is not None" there, but it's not critical. I like your solution:)

VadimPusiak788 (approved):

PR #3047, opened 2026-08-11
Solution

nastiagoncharova910-spec:


app/main.py
N1ckw3ll, line 1: The code works; I'll approve it. But the bot probably won't, because there's a checklist of requirements or recommendations.
3. When creating a list of instances from a collection (such as a list of dictionaries), it is  recommended to use **list comprehension** rather than traditional `for` loops. 
4. When accessing values in a dictionary, it is better to use the `dict.get()` method instead of explicitly checking for the presence of a key. This approach simplifies the code and avoids potential `KeyError` exceptions, making it cleaner and more concise.

N1ckw3ll (approved):

